PR #57, opened 2022-07-18
update cvp script to output inv file

ksator:
update cvp script to output an inv file for the inv module

### DIFF
--- a/scripts/create-devices-inventory-from-cvp.py
+++ b/scripts/create-devices-inventory-from-cvp.py
@@ -15,12 +15,26 @@ import json
 import os
 from argparse import ArgumentParser
 from getpass import getpass
+import yaml
 import requests
 from cvprac.cvp_client import CvpClient
+from anta.inventory import AntaInventory
 
 # Ignore certificate warnings
 # pylint: disable=E1101
 requests.packages.urllib3.disable_warnings()
+
+def create_inventory(cvp_inventory, out_dir, container):
+    """
+    create an inventory file
+    """
+    inv={AntaInventory.INVENTORY_ROOT_KEY:{"hosts":[]}}
+    for dev in cvp_inventory:
+        inv[AntaInventory.INVENTORY_ROOT_KEY]['hosts'].append({"host":dev["ipAddress"]})
+    # write the devices IP address in a file
+    out_file = f'{out_dir}/{container}.yml'
+    with open(out_file, 'w', encoding="utf8") as out_file:
+        out_file.write(yaml.dump(inv))
 
 def main():
     parser = ArgumentParser(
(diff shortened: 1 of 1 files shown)


scripts/create-devices-inventory-from-cvp.py
```diff
@@ -81,23 +81,30 @@ def main():
...
         # Get a list of all devices
-        # help(clnt.api.get_inventory)
         inventory = clnt.api.get_inventory()
-        # write all IP addresses from the CVP inventory in the all.txt file
-        out_file = f'{out_dir}/all.txt'
+        # write the IP addresses in a file
+        out_file = f'{out_dir}/all.yml'
         with open(out_file, 'w', encoding="utf8") as out_file:
+            out_file.write("anta_inventory:")
+            out_file.write('\n')
```
titom73, line 89: Why not having `\n` in the previous line ?

scripts/create-devices-inventory-from-cvp.py
```diff
@@ -81,23 +81,30 @@ def main():
...
+            out_file.write("  hosts:")
+            out_file.write('\n')
             for dev in inventory:
```
titom73, line 92: Would be nice to use model we have for inventory to capture CVP data and then save it instead of hardcoding inventory syntax.

Below is a pseudo code that could help:

```python
cvp_data[anta.inventory.AntaInventory.INVENTORY_ROOT_KEY]['hosts'] = []
for dev in container_inventory:
    cvp_data[anta.inventory.AntaInventory.INVENTORY_ROOT_KEY]['hosts'].append(dev['ipAddress'])

with open(cvp_inventory, 'w') as f:
    yaml.dump(cvp_data, f)
```

titom73 (changes requested):

scripts/create-devices-inventory-from-cvp.py
```diff
@@ -81,23 +81,30 @@ def main():
...
-        # Get devices under a container and write the devices IP address in a file
-        # help(clnt.api.get_devices_in_container)
+        # Get devices under a container
         container = args.container
         container_inventory = clnt.api.get_devices_in_container(container)
-        out_file = f'{out_dir}/{container}.txt'
+        # write the devices IP address in a file
+        out_file = f'{out_dir}/{container}.yml'
         with open(out_file, 'w', encoding="utf8") as out_file:
+            out_file.write("anta_inventory:")
```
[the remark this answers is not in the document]
ksator: OK I will use it

scripts/create-devices-inventory-from-cvp.py
ksator: I will use the model from the inventory module

scripts/create-devices-inventory-from-cvp.py
ksator: I will use the model from the inventory module

ksator: will looks like this
```
from anta.inventory import AntaInventory
import yaml

inv={AntaInventory.INVENTORY_ROOT_KEY:{"hosts":[]}}

cvp=["1.1.1.1", '2.2.2.22']

for dev in cvp: 
   inv[AntaInventory.INVENTORY_ROOT_KEY]['hosts'].append({"host":dev})

with open("inventory.yml","w", encoding="utf8") as outfile 
    outfile.write(yaml.dump(inv))
```

PR #65, opened 2022-07-20
feat(inventory): Limit discovery threads to 100 if inventory is bigger than 100

titom73:


anta/inventory/__init__.py
```diff
@@ -471,8 +473,9 @@ def refresh_device_facts(self) -> None:
         Execute in parallel a call to _refresh_online_flag_device to test device connectivity.
         """
         logger.debug(f'Refreshing facts for current inventory')
-        number_of_devices = len([dev.host for dev in self._inventory])
-        with Pool(processes=number_of_devices) as pool:
+        number_of_threads = len(self._inventory) if len(
+            self._inventory) < self.MAX_DISCOVERY_THREADS else self.MAX_DISCOVERY_THREADS
+        with Pool(processes=number_of_threads) as pool:
```
mtache, line 478: By default, CPU count is used so no need to have an hardcoded maximum value IMHO (especially 100 that will spawn 100 processes on a 16 cores machine).
Also, any reason why processes are used rather than threads here ? I do not see any CPU bound/CPU intensive work, just I/O wait so spawning processes is probably overkill :)

mtache: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool

ksator (changes requested):
we are importing twice multiprocessing
```
import multiprocessing
from multiprocessing import Pool
multiprocessing.cpu_count
with Pool
```
is it not better to do that?
```
import multiprocessing
multiprocessing.cpu_count
with multiprocessing.Pool

PR #67, opened 2022-07-20
doc: Update to reflect inventory format

titom73:
Update README file to use new inventory format as most of the scripts have been converted now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,11 +69,22 @@ management api http-commands
 
 First, we need to list devices we want to test. You can create a file manually with this format:
 
-```txt
-10.73.1.101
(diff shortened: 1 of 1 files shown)


ksator (changes requested):
we can use 
network: 192.168.110.0/24 
no need to use 
network: '192.168.110.0/24'

titom73:
> we can use network: 192.168.110.0/24 no need to use network: '192.168.110.0/24'

Both are identical as from YAML perspective, it is a string. Also, having quote means this value will be evaluated as a string in any case.

ksator (approved):

PR #77, opened 2022-07-22
refactor(tests): Update MULTICAST tests to support new structures

titom73:


anta/tests/multicast.py
```diff
@@ -1,60 +1,84 @@
...
+            "verify_igmp_snooping_global was not run as no configuration was givem"
+        )
+        return result
     try:
-        if response[0]['igmpSnoopingState'] == configuration:
-            return True
-        return False
-    except KeyError:
-        return None
+        response = device.runCmds(1, ["show ip igmp snooping"], "json")
```
gmuloc, line 74:
```suggestion
        response = device.session.runCmds(1, ["show ip igmp snooping"], "json")
```

PR #98, opened 2022-07-28
refactor(tests): Update MLAG tests with logging

titom73:


titom73:
Exception is not caugth correctly

titom73:
> Exception is not caugth correctly

Fixed

PR #107, opened 2022-08-30
feat(script): Update check-devices.py to use ANTA objects

titom73:


scripts/check-devices.py
```diff
@@ -1,186 +1,193 @@
...
+    ############################################################################
+    # Test Execution
+    ############################################################################
+
+    logger.info('starting running test on by_host ...')
+    manager = ResultManager()
+    list_tests = []
+    for device, test in itertools.product(inventory_anta.get_inventory(), tests_catalog):
+        if (cli_options.hostip is None or cli_options.hostip == str(device.host)) and \
+           (cli_options.test is None or cli_options.test == str(test[0].__name__)):
```
gmuloc, line 160:
```suggestion
        if ((cli_options.hostip is None or cli_options.hostip == str(device.host)) and
            (cli_options.test is None or cli_options.test == str(test[0].__name__))):
```

scripts/create-devices-inventory-from-cvp.py
```diff
@@ -68,7 +68,7 @@ def main() -> None:
     payload = json.dumps({"userId": args.username, "password": args.password})
     headers = {"Content-Type": "application/json", "Accept": "application/json"}
     response = requests.request(
-        "POST", URL, headers=headers, data=payload, verify=False
+        "POST", URL, headers=headers, data=payload, verify=False, timeout=10
```
gmuloc, line 71: Probably should catch timeout error here with some "meaningful" message?

ksator:
Could we please rename `--by-test-cases` with `--by-test`?

scripts/create-devices-inventory-from-cvp.py
titom73: I would let @ksator updating his script :)

This update is only because linting complained about this missing parameter in the CI

titom73:
> Could we please rename `--by-test-cases` with `--by-test`?

Change done in script and documentation

scripts/create-devices-inventory-from-cvp.py
ksator: OK I will do this later on

ksator (approved):
LGTM

Issue #118, opened 2022-09-15
Add support for custom defined tests


titom73:
## Description

In order to allow users to build their own test collection, it would be nice to add option to support catalog pointing to custom libraries.

Today, we only look at `anta.tests` package but we could extend YAML format to support both `anta.tests` and a custom package with the following format:

```yaml
<package.name>:
      - <test function>:
           <function options>
```

An example would be:

```yaml
anta.tests.software:
  - verify_eos_version: # Verifies the device is running one of the allowed EOS version.
      versions: # List of allowed EOS versions.
        - 4.25.4M
        - 4.26.1F
inetsix.tests.mydesign:
  - check_topology:
```

## Potential impacts:

- `anta.loader.parse_catalog`

PR #119, opened 2022-09-16
feat(anta.loader)!: Add support for path to custom tests

titom73:
Add ability to load user's defined tests using ANTA catalog loader (Issue #118 )

Now, catalog uses python module name to target correct tests:

```yaml
---
anta.tests.software:
  - verify_eos_version:
      versions: 
        - 4.25.4M
        - 4.26.1F
anta.tests.routing:
  generic:
    - verify_routing_protocol_model:
        model: multi-agent
```

First level has to use full module name and then, loader do recursive search within this module to find correct submodules and tests

## Breaking changes

- Catalog format __must__ use `anta.tests.<test module>` and not just `<test module>`

## Todo

- [x] Validate behaviour with external test library
- [x] Update documentation for catalog
- [ ] Document how to create custom tests lib


ksator (approved):
LGTM

PR #121, opened 2022-09-19
feat(anta.inventory): Add support for DNS names, eAPI port and device alias names

mtache:
When using a containerlab topology on macOS, we need to use port forwarding to localhost to access eAPI endpoints.
The current inventory only supports IP addresses. The PR is to support hostnames (DNS names instead of IP), ports and alias names in the inventory.

The following inventory works with this PR:
```
anta_inventory:
  hosts:
  - name: spine1
    host: localhost
    port: 44301
  - name: spine2
    host: localhost
    port: 44302
  - name: leaf1
    host: localhost
    port: 44311
  - name: leaf1
    host: localhost
    port: 44312
  - name: leaf3
    host: localhost
    port: 44313
  - name: leaf4
    host: localhost
    port: 44314
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,27 +1,47 @@
 {
-    "version": "0.2.0",
-    "configurations": [
-        {
-            "name": "check-devices debug",
-            "type": "python",
-            "request": "launch",
-            "program": "${workspaceFolder}/scripts/check-devices.py",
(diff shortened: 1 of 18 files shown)


mtache:
Closing this pull request as the eAPIUrl model is not used anymore with asyncio framework

PR #122, opened 2022-09-19
feat(anta.inventory): use aioeapi to handle a device session 

mtache:
Using asyncio framework will help improving performances when running tests on a large network with a high number of tests.
First step is to refactor the anta.inventory module to use an async library, this PR tries an attempt with [aio-eapi](https://github.com/jeremyschulman/aio-eapi).
This PR must also refactor all the tests as coroutines and modify the scripts to run the asyncio event loop.

---

### Modifications to `anta.inventory`

The session attribute of InventoryDevice is now a aioeapi.Device object.
auto_connect is removed from the constructor: this would cause to run an asyncio event loop in an object constructor and prevent fine control from a ANTA user script on the asyncio scheduling.

`_is_device_online()` uses `aioeapi.Device.check_connection()` to only tries to open a port on the device.
`_ refresh_device_fact()` updates all the device flags and reduce the number of API calls compared to the previous implementation.

`connect_inventory()` is a coroutine that needs to be called in a ANTA user script to establish connection towards the devices.

```
async def main():
   await anta_inventory.connect_inventory()
```

---

When using a containerlab topology on macOS, we need to use port forwarding to localhost to access eAPI endpoints.
The current inventory only supports IP addresses. This PR supports hostnames (DNS names instead of IP), ports and alias names in the inventory.

The following inventory works with this PR:

```
anta_inventory:
  hosts:
  - name: spine1
    host: localhost
    port: 44301
  - name: spine2
    host: localhost
    port: 44302
  - name: leaf1
    host: localhost
    port: 44311
  - name: leaf1
    host: localhost
    port: 44312
  - name: leaf3
    host: localhost
    port: 44313
  - name: leaf4
    host: localhost
    port: 44314
 ```

### DIFF
--- a/.github/scripts/anta-tester.py
+++ b/.github/scripts/anta-tester.py
@@ -59,7 +59,7 @@ logging.getLogger('anta.tests.routing.generic').setLevel(logging.ERROR)
 logging.getLogger('anta.tests.routing.bgp').setLevel(logging.ERROR)
 logging.getLogger('anta.tests.routing.ospf').setLevel(logging.ERROR)
 
-async def main(inventory_anta: AntaInventory, tests_catalog: List[Tuple[Callable[..., TestResult], Dict[Any, Any]]]):
+async def main(inventory_anta: AntaInventory, tests_catalog: List[Tuple[Callable[..., TestResult], Dict[Any, Any]]]) -> None:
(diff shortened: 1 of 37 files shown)


gmuloc (approved):
LGTM but someone should try it on some env ;)

PR #123, opened 2022-09-20
Feat: add anta_test decorator

gmuloc:


gmuloc:
@titom73 / @mtache - this PR may be a bit black magic for users - the decorator is injecting the `Result` object and so you can call the test function without the TestResult argument.
It may be better to have the result object as an "optional" keyword argument injected by the decorator so it does not seem too voodoo.

or maybe another solution? 

Thoughts ?

mtache:
Ship it !

mtache (approved):

PR #127, opened 2023-01-06
feat(asyncio): update scripts

mtache:
Update scripts to work with the new ANTA asyncio framework.


scripts/clear-counters.py
```diff
@@ -4,87 +4,106 @@
...
+    await inv.connect_inventory()
+    devices = inv.get_inventory(established_only=True)
     for device in devices:
-        switch = device.session  # type: ignore
-        host = str(device.host)  # type: ignore
         try:
-            if device.hw_model in ["cEOSLab", "vEOS-lab"]:  # type: ignore
-                switch.runCmds(
-                    1, [{"cmd": "enable", "input": enable_pass}, "clear counters"]
+            if device.hw_model in ["cEOSLab", "vEOS-lab"]:
```
gmuloc, line 51: I would rewrite this to build first the line of commands based on the device hw_model and then call only once await + logger

Issue #128, opened 2023-01-11
[Bug] - check-device.py fails with --by-host option


titom73:
The `--by-host` knob is broken and returns a trace:

```bash
$ check-devices.py -i anta-inventory.yml -c tests-bases.yml -u xxx -p xxx --by-host --table

Traceback (most recent call last):
  File "/home/tom/.pyenv/versions/arista-avd-3.9/bin/check-devices.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/tom/Projects/arista/network-test-automation/scripts/check-devices.py", line 312, in <module>
    reporter.report_summary_hosts(
  File "/home/tom/Projects/arista/network-test-automation/anta/reporter/__init__.py", line 251, in report_summary_hosts
    for host_read in result_manager.get_hosts():
  File "/home/tom/Projects/arista/network-test-automation/anta/result_manager/__init__.py", line 197, in get_hosts
    if str(testcase.host) not in result_list:
AttributeError: 'TestResult' object has no attribute 'host'
```

This option was working in v0.3.0 and is a regression part of one of these PRs:

- 85dca38 (HEAD -> master, upstream/master, origin/master, origin/HEAD) Doc: Update test pattern in documentation (#126)
- 630bff2 feat(asyncio): update scripts (#127)
- 6d50ac7 feat(anta.inventory): use aioeapi to handle a device session  (#122)
- 5b16412 Feat: add anta_test decorator (#123)
- cbc6091 feat(anta.loader)!: Add support for path to custom tests (#119)

gmuloc:
Closed by https://github.com/arista-netdevops-community/network-test-automation/pull/129

PR #129, opened 2023-01-11
fix: Rename field in get_hosts() to match updated TestResult model

titom73:
Fixing issue #128

- rename `testcase.host` by `testcase.name` in `get_hosts()` 
- rename `result.host` by `result.name` in`get_result_by_host()`


gmuloc (approved):
we are agreed

PR #132, opened 2023-01-12
feat(check-devices): Update default logging behavior

titom73:
Update default logging behaviour to not list all tests addition in result manager by default

- Configure `INFO` only for `check-devices.py` script
- Configure `WARNING` for `anta.inventory` module
- Configure `WARNING` for `anta.result_manager` module

To change logging for `anta.inventory` & `anta.result_manager`, `--loglevel` option can be used

### DIFF
--- a/scripts/check-devices.py
+++ b/scripts/check-devices.py
@@ -50,8 +50,12 @@ def setup_logging(level: str) -> None:
     * anta.result_manager
     * check-devices
 
+    By default, configure INFO for the script,
+    WARNING for anta.inventory and anta.report_manager
+    and CRITICAL for others
+
     Args:
-        level (str, optional): level name to configure. Defaults to 'critical'.
+        level (str): level name to configure.
     """
 
     if level is not None:
@@ -222,7 +226,7 @@ def cli_manager() -> argparse.Namespace:
     parser.add_argument(
         "-log",
         "--loglevel",
-        help="Provide logging level. Example --loglevel debug, default=none",
+        help="Provide logging level. Example --loglevel debug, By default INFO only for the script, others are configured to WARNING and CRITICAL",
     )
 
     return parser.parse_args()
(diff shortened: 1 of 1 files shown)


scripts/check-devices.py
```diff
@@ -214,7 +222,7 @@ def cli_manager() -> argparse.Namespace:
     parser.add_argument(
         "-log",
         "--loglevel",
-        default="info",
+        default="unset",
```
gmuloc, line 225: Can we just remove the default value and test if the value is set or not?

titom73: Fixed with last commit

scripts/check-devices.py
```diff
@@ -41,7 +41,7 @@
 logger = logging.getLogger(__name__)
 
 
-def setup_logging(level: str = "info") -> None:
+def setup_logging(level: str) -> None:
```
gmuloc, line 44: Need to change the Args now to `level (str)` and maybe mention tht the default is actually `WARNING` for some logger and INFO for the main one ;)

scripts/check-devices.py
```diff
@@ -214,8 +222,7 @@ def cli_manager() -> argparse.Namespace:
     parser.add_argument(
         "-log",
         "--loglevel",
-        default="info",
-        help="Provide logging level. Example --loglevel debug, default=info",
+        help="Provide logging level. Example --loglevel debug, default=none",
```
gmuloc, line 225: Maybe let's explain add the default behavior here as well

scripts/check-devices.py
titom73: done

scripts/check-devices.py
titom73: Done with a short description to not make output not readable

gmuloc (approved):
LGTM - though people may wonder what "other" loggers are which are configured to WARNING and CRITICAL by default :) (and looks like most of them are actually ERROR no?)

PR #133, opened 2023-01-25
feat(cli): Initial implementation of ANTA CLI

titom73:
Implement ANTA cli using [click()](https://click.palletsprojects.com/en/8.1.x/)

## Basic Usage

```bash
anta
Usage: anta [OPTIONS] COMMAND [ARGS]...

  Arista Network Test CLI

Options:
  --username TEXT         Username to connect to EOS  [env var: ANTA_USERNAME;
                          required]
  --password TEXT         Password to connect to EOS  [env var: ANTA_PASSWORD;
                          required]
  --timeout INTEGER       Connection timeout (default 5)  [env var:
                          ANTA_TIMEOUT]
  --enable-password TEXT  Enable password if required to connect  [env var:
                          ANTA_ENABLE_PASSWORD]
  -i, --inventory PATH    Path to your inventory file  [env var:
                          ANTA_INVENTORY; required]
  --help                  Show this message and exit.

Commands:
  exec  Execute commands to inventory devices
  get   Get data from/to ANTA
  nrfu  Run NRFU against inventory devices
```

Full CLI description: [Repository documentation / mkdocs](https://github.com/titom73/network-test-automation/blob/features/cli/docs/usage-anta-cli.md)


titom73:
Documentation will come once we agree on initial cli suggestion.

anta/cli/check/commands.py
```diff
@@ -0,0 +1,102 @@
...
+logger = logging.getLogger(__name__)
+
+
+@anta.command(no_args_is_help=True)
+@click.pass_context
+# Generic options
+@click.option('--inventory', '-i', show_envvar=True, prompt='Inventory path', help='Path to your inventory file', type=click.Path())
+@click.option('--catalog', '-c', show_envvar=True, prompt='Path for tests catalog', help='Path for tests catalog', type=click.Path())
+@click.option('--tags', '-t', default='all', help='List of tags using coma as separator: tag1,tag2,tag3', type=str)
+@click.option('--display', default='table', type=click.Choice(['table', 'json', 'list'], case_sensitive=False), help='output format selection. default is table')
```
gmuloc: Can we find a way via click to avoid repeating the options for both command (like some kind of "group of options" that could be apploed to both)?

anta/cli/check/utils.py
```diff
@@ -0,0 +1,107 @@
...
+        password=password,
+        enable_password=enable_password,
+        timeout=timeout
+    )
+    logger.info(f"Inventory {inventory} loaded")
+
+    ############################################################################
+    # Test loader
+    ############################################################################
+
```
gmuloc, line 47:
```suggestion
    # Test loader
```

anta/cli/check/utils.py
```diff
@@ -0,0 +1,107 @@
...
+    return results
+
+
+def display_table(console: Console, results: ResultManager, group_by: str = 'none', search: str = '') -> None:
+    """Display result in a table"""
+    reporter = ReportTable()
+    if group_by == 'none':
+        console.print(
+            reporter.report_all(result_manager=results)
+        )
```
gmuloc, line 76:
```suggestion
def display_table(console: Console, results: ResultManager, group_by: Optional[str] = None , search: str = '') -> None:
    """Display result in a table"""
    reporter = ReportTable()
    if group_by is None:
        console.print(
            reporter.report_all(result_manager=results)
        )
```

anta/cli/check/utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def display_json(console: Console, results: ResultManager, output_file: str = '') -> None:
+    """Display result in a json format"""
+    console.print(Panel("JSON results of all tests", style="cyan"))
+    print_json(results.get_results(output_format="json"))
+    if output_file != '':
+        with open(output_file, "w", encoding="utf-8") as fout:
+            fout.write(results.get_results(output_format="json"))
```
gmuloc, line 98:
```suggestion
def display_json(console: Console, results: ResultManager, output_file: Optional[str] = None) -> None:
    """Display result in a json format"""
    console.print(Panel("JSON results of all tests", style="cyan"))
    print_json(results.get_results(output_format="json"))
    if output_file is not None:
        with open(output_file, "w", encoding="utf-8") as fout:
            fout.write(results.get_results(output_format="json"))
```

Should we print even when outputting to file?

anta/cli/check/utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def display_list(console: Console, results: ResultManager, output_file: str = '') -> None:
```
gmuloc, line 101: same as above for default value for output_file

anta/cli/check/utils.py
```diff
@@ -0,0 +1,107 @@
...
+# pylint: disable=cyclic-import
+# pylint: disable=R0401
+
+"""
+Utils functions to use with anta.cli.check.commands module.
+"""
+
+import asyncio
+import logging
+from typing import Any
```
gmuloc, line 13: Add Optional here

anta/cli/check/utils.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+# pylint: disable=too-many-arguments
+# pylint: disable=cyclic-import
```
gmuloc, line 4: This should be fixed what is the issue? if it is about typing here is the correct pattern: https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports 

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from main import Main
```

anta/cli/check/utils.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+# pylint: disable=too-many-arguments
```
gmuloc, line 3: Better to add this at the function level rather than top of file

anta/cli/check/utils.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/python
```
gmuloc, line 1: https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take 

I would rather we don't include it as we don't really need the file to be executable but better to use `/usr/bin/env python`

gmuloc (commented):
have not gone through everything but some thoughts and suggestions

anta/cli/check/commands.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+    )
+
+    results = check_run(
+        inventory=inventory,
+        catalog=catalog,
+        username=ctx.obj['username'],
+        password=ctx.obj['password'],
+        timeout=ctx.obj['timeout'],
+        enable_password='unset',
```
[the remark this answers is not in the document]
titom73: Actually, it should be `enable_password=ctx.obj['enable_password']` since enable_password is defined in cli

Going to be fixed in next commit

anta/cli/check/commands.py
titom73: - Have not seen such option in Click or its plugins.
- So I have moved inventory under cli like username and password.
- Cagtalog and display are not for all entrypoint so it should not be changed.
- Tags is still present at this level as it is probably the most updated element to run CLI

mtache (approved):
Ship it!

anta/cli/check/commands.py
```diff
@@ -0,0 +1,133 @@
...
+from .utils import check_run, display_table, display_json
+
+logger = logging.getLogger(__name__)
+
+
+@click.command(no_args_is_help=True)
+@click.pass_context
+# Generic options
+@click.option('--catalog', '-c', show_envvar=True, prompt='Path for tests catalog', help='Path for tests catalog', type=click.Path(), required=True)
+@click.option('--tags', '-t', default='all', help='List of tags using coma as separator: tag1,tag2,tag3', type=str, required=False)
```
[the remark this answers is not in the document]
titom73: Fixed in latest commit

PR #137, opened 2023-02-09
Chore: Move data to pyproject.toml

gmuloc:
* Prepare for upload to PIP
* Version bumped to 0.4.0 for now
* Simplify tox run (run linting and typing only once)


titom73:
Looks good using `pip install .` and `tox -e lint`

But, we cannot install using `-e` option to be in editable during development process

```bash
root@9cdd90febc39:/tmp# pip list | grep setuptools
setuptools         67.2.0

root@9cdd90febc39:/tmp# pip install -e .
Obtaining file:///tmp
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
ERROR: Project file:///tmp has a 'pyproject.toml' and its build backend is missing the 'build_editable' hook. 
Since it does not have a 'setup.py' nor a 'setup.cfg', it cannot be installed in editable mode. 
Consider using a build backend that supports PEP 660.
```

titom73 (approved):
LGTM!

PR #147, opened 2023-03-02
fix(anta): Support for implicit default tag when not set by user

titom73:
Today, when a tag is set in inventory, implicit default tag is not used and it is required for user to set and use a tag to cover all devices

This PR activates support for implicit default `all` tag meaning if CLI does not specify a tag, then command is run against the entire inventory.

Issue initially reported by @aphillipps 

Also, it adds a new CLI entrypoint to display inventory and help troubleshooting

```bash
$ anta get inventory --help
Usage: anta get inventory [OPTIONS]

  Show inventory loaded in ANTA.

Options:
  -t, --tags TEXT                 List of tags using comma as separator:
                                  tag1,tag2,tag3
  --log-level, --log [debug|info|warning|critical]
                                  Logging level of the command
  --connected / --not-connected   Display inventory after connection has been
                                  created
  --help                          Show this message and exit.

$ anta --inventory .personal/inventory_atd.yml get inventory -t spine --log debug
Current inventory content is:
[
  {
    "name": "spine01",
    "host": "192.168.0.10",
    "username": "ansible",
    "password": "ansible",
    "port": "443",
    "enable_password": "None",
    "session": "<aioeapi.device.Device object at 0x7fc272c9a4f0>",
    "hw_model": "unset",
    "tags": "['fabric', 'spine', 'all']",
    "timeout": "10.0",
    "established": "False",
    "is_online": "False"
  },
...
```

Also a command to list configured tags

```bash
anta --inventory .personal/inventory_atd.yml get tags --log debug
Tags found:
[
  "all",
  "fabric",
  "leaf",
  "spine"
]
None

* note that tag all has been added by anta
```


gmuloc (commented):
Hmmm I may be missing something but it looks like there is no way to disable the default tag in this PR - so can you still filter on tags after this?

anta/inventory/__init__.py
```diff
@@ -304,18 +305,19 @@ def get_inventory(
...
-            tags (List[str], optional): List of tags to use to filter devices. Default is [default].
+            tags (List[str], optional): List of tags to use to filter devices.
 
         Returns:
             InventoryDevices: An inventory with concerned devices
         """
-        if tags is None:
-            tags = [DEFAULT_TAG]
 
         inventory_filtered_tags = InventoryDevices()
```
gmuloc, line 314:
```suggestion

        def _filter_devices(device):
            """
            Helper function to select the devices based on the input tags
            and the requirement for an established connection.
            """
            #  is the device filtered by tags
            if tags is not None and not(any(tag in tags for tag in device.tags)):
              return False
            # is the device filtered because session was not established
            if established_only and not device.established:
              return False
            return True
            
        return InventoryDevices(list(filter(_filter_devices, self._inventory)))
```

could replace the `any` with some set intersection even.

titom73: Yes, but I think it is not required by the code anymore. And this code block is not here in the last commit AFAIK.

gmuloc: thats a replacement for the logic of the whole function - github UI did not let me make it nice

titom73: Ok, It is now implemented with some small arrangements

anta/cli/get/commands.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python
+# coding: utf-8 -*-
+# pylint: disable = redefined-outer-name
```
gmuloc, line 3: Can we get rid of this pylint ignore?

gmuloc (approved):
LGTM

anta/cli/get/commands.py
titom73: Not easily and making code more complex:

It is related because we have an option named `tags` in `anta get inventory` and we also have a function named `tags` for `anta get tags`

PR #148, opened 2023-03-06
fix(anta.tests): Allow verify_running_config to support no enable pasword

titom73:
Allow to run `verify_running_config` whitout setting a fake enable password if not configured in CLI.

### DIFF
--- a/anta/tests/configuration.py
+++ b/anta/tests/configuration.py
@@ -57,15 +57,24 @@ async def verify_running_config_diffs(
         * result = "error" if any exception is caught
 
     """
-    device.assert_enable_password_is_not_none("verify_running_config_diffs")
-
-    response = await device.session.cli(
-        commands=[
-            {"cmd": "enable", "input": str(device.enable_password)},
-            "show running-config diffs",
-        ],
-        ofmt="text"
-    )
+    if device.enable_password is not None:
+        device.assert_enable_password_is_not_none(
+            "verify_running_config_diffs")
+        response = await device.session.cli(
+            commands=[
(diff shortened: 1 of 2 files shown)


anta/tests/configuration.py
```diff
@@ -57,15 +57,24 @@ async def verify_running_config_diffs(
...
-
-    response = await device.session.cli(
-        commands=[
-            {"cmd": "enable", "input": str(device.enable_password)},
-            "show running-config diffs",
-        ],
-        ofmt="text"
-    )
+    if device.enable_password is not None:
+        device.assert_enable_password_is_not_none(
```
gmuloc, line 61: Remove the assertion as you have already checked line 60 that the password was not None

titom73: Done. Thanks for the comment

anta/tests/configuration.py
```diff
@@ -57,15 +57,22 @@ async def verify_running_config_diffs(
...
+            ofmt="text"
+        )
+    else:
+        response = await device.session.cli(
+            commands=[
+                {"cmd": "enable"},
+                "show running-config diffs",
+            ],
+            ofmt="text"
+        )
```
gmuloc, line 75:
```suggestion
    if device.enable_password is not None:
        enable_cmd = {"cmd": "enable", "input": str(device.enable_password)}
    else:
         enable_cmd = "enable"
    commands = [enable_cmd, "show running-config diffs"]
    response = await device.session.cli(
        commands=commands,
        ofmt="text",
    )
```

PR #150, opened 2023-03-07
doc: Update structure & Add new commands

titom73:


docs/cli/exec.md
```diff
@@ -0,0 +1,55 @@
...
+  -t, --tags TEXT                 List of tags using comma as separator:
+                                  tag1,tag2,tag3
+  --log-level, --log [debug|info|warning|critical]
+                                  Logging level of the command
+  --help                          Show this message and exit.
+```
+
+## Collect a set of commands
+
+This command collects all commands you defined in a catalog. It can bei either `json` or `text`.
```
gmuloc, line 25:
```suggestion
This command collects all commands you defined in a catalog. It can be either `json` or `text`.
```

docs/cli/get-inventory-information.md
```diff
@@ -0,0 +1,159 @@
+# Gest Inventory Information
```
gmuloc, line 1:
```suggestion
# Get Inventory Information
```

docs/cli/get-inventory-information.md
```diff
@@ -0,0 +1,159 @@
...
+
+Options:
+  --log-level, --log [debug|info|warning|critical]
+                                  Logging level of the command
+  --help                          Show this message and exit.
+```
+
+### Example
+
+Let consider the following inventory:
```
gmuloc, line 25:
```suggestion
Let's consider the following inventory:
```

docs/cli/get-inventory-information.md
```diff
@@ -0,0 +1,159 @@
...
+    tags: ['fabric', 'leaf']
+  - host: 192.168.0.14
+    name: leaf03
+    tags: ['fabric', 'leaf']
+  - host: 192.168.0.15
+    name: leaf04
+    tags: ['fabric', 'leaf']
+```
+
+To get list of all configured tags in your CLI, run the following command:
```
gmuloc, line 50:
```suggestion
To get the list of all configured tags in your CLI, run the following command:
```

docs/cli/get-inventory-information.md
```diff
@@ -0,0 +1,159 @@
...
+```bash
+$ anta get tags
+Tags found:
+[
+  "all",
+  "fabric",
+  "leaf",
+  "spine"
+]
+None
```
gmuloc, line 61: C'est normal le print None ici?

docs/cli/get-inventory-information.md
```diff
@@ -0,0 +1,159 @@
...
+
+* note that tag all has been added by anta
```
gmuloc, line 63:
```suggestion
* note that tag 'all' has been added by anta
```

docs/cli/get-inventory-information.md
```diff
@@ -0,0 +1,159 @@
...
+
+!!! info Default tag
+    As you can see, the tag `all` has been added even if not explicitely configued in your inventory. This tag is the default tag added to all your devices to run commands against your inventory when you do not provide any specific tag.
+
+## List devices in inventory
+
+### Command overview
+
+To get a list of all devices available in your inventory with ANTA information, you can simply use following command
+
```
gmuloc, line 74:
```suggestion
To get a list of all devices available in your inventory with ANTA, use the following command

```

docs/cli/get-inventory-information.md
```diff
@@ -0,0 +1,159 @@
...
+                                  Logging level of the command
+  --connected / --not-connected   Display inventory after connection has been
+                                  created
+  --help                          Show this message and exit.
+```
+
+It will give you all information loaded in ANTA inventory from your [inventory file](../../usage-inventory-catalog/).
+
+!!! info Offline information only
+    By default information are only those not based on device connection. If you want to get information based on connection such as hardware, you should use `--connected` option.
```
gmuloc, line 94:
```suggestion
    By default only information not based on device connection is available. If you want to get information based on connection such as hardware model, you should use the `--connected` option.
```

docs/cli/get-inventory-information.md
```diff
@@ -0,0 +1,159 @@
...
+    tags: ['fabric', 'leaf']
+  - host: 192.168.0.14
+    name: leaf03
+    tags: ['fabric', 'leaf']
+  - host: 192.168.0.15
+    name: leaf04
+    tags: ['fabric', 'leaf']
+```
+
+You can get ANTA inventory with command:
```
gmuloc, line 123:
```suggestion
You can get ANTA inventory with the command:
```

docs/cli/inv-from-cvp.md
```diff
@@ -0,0 +1,38 @@
+# Create inventory from CloudVision
+
+In large setup, it can be useful to create your inventory based on CloudVision inventory.
```
gmuloc, line 3:
```suggestion
In a large setup, it can be useful to create your inventory based on CloudVision inventory.
```

docs/cli/inv-from-cvp.md
```diff
@@ -0,0 +1,38 @@
...
+  -u, --cvp-username TEXT         CVP Username
+  -p, --cvp-password TEXT         CVP Password / token
+  -c, --cvp-container TEXT        Container where devices are configured
+  -d, --inventory-directory PATH  Path to save inventory file
+  --log-level, --log [debug|info|warning|critical]
+                                  Logging level of the command
+  --help                          Show this message and exit.
+```
+
+Output is an inventory with name of the container added as a tag for the host
```
gmuloc, line 22:
```suggestion
Output is an inventory with the name of the container added as a tag for the host:
```

docs/cli/inv-from-cvp.md
```diff
@@ -0,0 +1,38 @@
...
+    tags:
+    - pod1
+  - host: 192.168.0.15
+    name: leaf4
+    tags:
+    - pod2
+```
+
+!!! warning Container lookup is not recursive
+    Current implementation only takes devices directly attached to a specific container when using cli with `--cvp-container` option.
```
gmuloc, line 38: do we have some tip if the user needs to get nested devices? (I guess run multiple time the command and append to some common inventory file?)

docs/cli/nrfu.md
```diff
@@ -0,0 +1,94 @@
+# Execute NRFU testing
+
+All the NRFU testing commands are placed under `anta nrfu` and provides different rendering options:
```
gmuloc, line 3:
```suggestion
All the NRFU testing commands are placed under `anta nrfu` and provide different rendering options:
```

docs/cli/nrfu.md
```diff
@@ -0,0 +1,94 @@
...
+leaf2 :: verify_portchannels :: SUCCESS
+leaf2 :: verify_illegal_lacp :: SUCCESS
+leaf2 :: verify_loopback_count :: FAILURE (Found 3 Loopbacks when expecting 2)
+leaf2 :: verify_svi :: SUCCESS
+[...]
+```
+
+## NRFU with table report
+
+This rendering print results in a nice table suppoting grep filtering. It comes with its own set of options:
```
gmuloc, line 59:
```suggestion
This rendering prints results in a nice table supporting grep filtering. It comes with its own set of options:
```

docs/cli/nrfu.md
```diff
@@ -0,0 +1,94 @@
...
+
+- Search (`--search`) for a pattern in hostname and test name.
```
gmuloc, line 61: in the `text` section you mention `regex` and here `pattern` for the `search` option, is it the same?

docs/cli/get-inventory-information.md
titom73: Maybe better with oneline code ?

```suggestion
* note that tag `all` has been added by anta
```

docs/cli/overview.md
```diff
@@ -0,0 +1,66 @@
+# CLI overview
+
+ANTA comes with a CLI to execute all the supported actions. If you want to build your own tool, you should visit [this page](../usage-as-python-lib/) where we describe how to use ANTA as a python library
+
+To invoke anta, open a shell window and then enter `anta`
```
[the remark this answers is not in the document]
titom73: Interesting point: what do we want for anta: lower case or upper case. Let's discuss that in the group maybe ?

docs/cli/nrfu.md
titom73: Yes it is the same. If not clear I can change `regex` by `regex pattern`

docs/cli/get-inventory-information.md
titom73: Yes, it is a code returned by Rich in its `print_json` function. Later we can see how to remove it in the CLI

docs/cli/get-inventory-information.md
gmuloc: could be - is this output part of the command or is it something added? (I am not sure as it is in the code block)

docs/cli/inv-from-cvp.md
titom73: Not yet, but it would be great to support recursive get from CVP API. But that should come in another PR focusing on this feature.

docs/cli/get-inventory-information.md
titom73: Correct, CLI output is not using quote but color instead which is not available in this context.

gmuloc (approved):
:ship:

PR #166, opened 2023-04-05
fix(anta.cli): Improve 'anta exec snapshot'

mtache:
- Modified `collect_commands()` logic to use `asyncio.gather()` instead of the synchronous logic with a for loop.
- The `--output-directory` option of the `anta exec snapshot` command could be None if not provided and created a folder with that name. Added a default value and a timestamp to avoid overriding previous command collections.


titom73 (approved):
LGTM

PR #190, opened 2023-05-08
feat(anta.tests): Add new various security test cases

carl-baillargeon:
Hello,

PR to merge new security test cases.

Thanks

### DIFF
--- a/anta/tests/security.py
+++ b/anta/tests/security.py
@@ -1,0 +1,345 @@
+"""
+Test functions related to the EOS various security settings
+"""
+from __future__ import annotations
+
+import logging
+from typing import Any, Dict, Optional, cast
+
+from anta.models import AntaTest, AntaTestCommand
+
+logger = logging.getLogger(__name__)
+
+
+class VerifySSHStatus(AntaTest):
+    """
+    Verifies if the SSHD agent is disabled in the default VRF.
+
+    Expected Results:
+        success: The test will pass if the SSHD agent is disabled in the default VRF.
+        failure: The test will fail if the SSHD agent is NOT disabled in the default VRF.
+    """
+
+    name = "VerifySSHStatus"
+    description = "Verifies if the SSHD agent is disabled in the default VRF."
+    categories = ["security"]
+    commands = [AntaTestCommand(command="show management ssh", ofmt="text")]
+
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """
+        Run VerifySSHStatus validation.
+        """
+
+        self.logger.debug(f"self.instance_commands is: {self.instance_commands}")
+        command_output = self.instance_commands[0].output
+        self.logger.debug(f"dataset is: {command_output}")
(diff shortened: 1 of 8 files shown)


tests/units/anta_tests/security/__init__.py
```diff
@@ -0,0 +1 @@
+"""Init security"""
```
gmuloc, line 1: Leave this empty

mkdocs.yml
```diff
@@ -154,6 +154,7 @@ nav:
     - System: api/tests.system.md
     - Software: api/tests.software.md
     - SNMP: api/tests.snmp.md
+    - Security: api/tests.security.md
     - Routing:
```
gmuloc, line 158: LEt's reorganize these via alphabetical order? (I know this is not only your PR .. just thinking about usability!)

anta/tests/security.py
```diff
@@ -0,0 +1,345 @@
...
+
+    Expected Results:
+        * success: The test will pass if the SSHD agent is disabled in the default VRF.
+        * failure: The test will fail if the SSHD agent is NOT disabled in the default VRF.
+    """
+
+    name = "VerifySSHStatus"
+    description = "Verifies if the SSHD agent is disabled in the default VRF."
+    categories = ["security"]
+    commands = [AntaTestCommand(command="show management ssh", ofmt="text")]
```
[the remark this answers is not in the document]
gmuloc: ok just checked and for now this yields "this is an unconverted command"...
can we put a comment above the `commands` stating so?


PS: amazing that telnet one is available in json and not ssh..

anta/tests/security.py
```diff
@@ -0,0 +1,345 @@
...
+logger = logging.getLogger(__name__)
+
+
+class VerifySSHStatus(AntaTest):
+    """
+    Verifies if the SSHD agent is disabled in the default VRF.
+
+    Expected Results:
+        * success: The test will pass if the SSHD agent is disabled in the default VRF.
+        * failure: The test will fail if the SSHD agent is NOT disabled in the default VRF.
```
gmuloc, line 20: wondering if this would not be worth adding a VRF keyword (where the default would be default) to be able to check for example management VRF (common use case I think)

gmuloc (commented):
Thanks a lot for these! a couple of comments, most non directly related to the PR

anta/tests/security.py
carl-baillargeon: I was wondering the same too. eAPI returns an error when it's disabled in a specified VRF though. Let me check what we can do.

gmuloc: so should we wait for an update here @carl-baillargeon or merge as it is?

carl-baillargeon: Merge as it for now. I will wait your PR for templates and modify accordingly. Thanks

gmuloc (approved):

PR #194, opened 2023-05-10
feat(anta.tests): Add new logging test cases

carl-baillargeon:
Hello,

Please review this PR for new logging test cases.

Thanks

### DIFF
--- a/anta/tests/logging.py
+++ b/anta/tests/logging.py
@@ -16,6 +16,8 @@ class VerifyLoggingPersistent(AntaTest):
     """
     Verifies if logging persistent is enabled and logs are saved in flash.
 
+    'show logging' does not support json output yet
+
     Expected Results:
         * success: The test will pass if logging persistent is enabled and logs are in flash.
         * failure: The test will fail if logging persistent is disabled or no logs are saved in flash.
@@ -43,10 +45,14 @@ class VerifyLoggingPersistent(AntaTest):
         pattern = r"-rw-\s+(\d+)"
         match = re.search(pattern, dir_flash_output)
 
-        if match and "disabled" not in log_output and int(match.group(1)) > 0:
-            self.result.is_success()
-        else:
-            self.result.is_failure("Persistent logging is disabled or no logs are saved in flash")
+        self.result.is_success()
+
+        if "disabled" in log_output:
+            self.result.is_failure("Persistent logging is disabled")
+            return
+
+        if not match or int(match.group(1)) == 0:
+            self.result.is_failure("No persistent logs are saved in flash")
 
 
 class VerifyLoggingSourceIntf(AntaTest):
@@ -148,7 +154,7 @@ class VerifyLoggingLogsGeneration(AntaTest):
     categories = ["logging"]
     commands = [
         AntaTestCommand(command="send log level informational message ANTA VerifyLoggingLogsGeneration validation"),
-        AntaTestCommand(command="show logging | tail", ofmt="text"),
+        AntaTestCommand(command="show logging informational last 30 seconds | grep ANTA", ofmt="text"),
     ]
 
     @AntaTest.anta_test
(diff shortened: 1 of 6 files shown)


anta/tests/logging.py
```diff
@@ -0,0 +1,297 @@
...
+        AntaTestCommand(command="dir flash:/persist/messages", ofmt="text"),
+    ]
+
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """
+        Run VerifyLoggingPersistent validation.
+        """
+        self.logger.debug(f"self.instance_commands is: {self.instance_commands}")
+        log_output = self.instance_commands[0].output
```
gmuloc, line 38: When you have `ofmt="text"` cast to `str`
```suggestion
        log_output = cast(str, self.instance_commands[0].output)
```

anta/tests/logging.py
```diff
@@ -0,0 +1,297 @@
...
+
+from anta.models import AntaTest, AntaTestCommand
+
+logger = logging.getLogger(__name__)
+
+
+class VerifyLoggingPersistent(AntaTest):
+    """
+    Verifies if logging persistent is enabled and logs are saved in flash.
+
```
gmuloc, line 18:
```suggestion
    Verifies if logging persistent is enabled and logs are saved in flash.

    'show logging' does not support json output yet
```

anta/tests/logging.py
```diff
@@ -0,0 +1,297 @@
...
+        self.logger.debug(f"dataset of log_output command is: {log_output}")
+        self.logger.debug(f"dataset of dir_flash_output command is: {dir_flash_output}")
+
+        pattern = r"-rw-\s+(\d+)"
+        match = re.search(pattern, dir_flash_output)
+
+        if match and "disabled" not in log_output and int(match.group(1)) > 0:
+            self.result.is_success()
+        else:
+            self.result.is_failure("Persistent logging is disabled or no logs are saved in flash")
```
gmuloc, line 49: so I would rather write it like this:
with an optional return if you want to stop after finding out that persistent logging is disabled.

Random question - is your test not able to catch false positive if the log file is actually 0 in size because no log has been captured yet?

```suggestion
        self.result.is_success()
        if "disabled" in log_output:
            self.result.is_failure("Persistent logging is disabled!")
            # return
        if not match or int(match.group(1)) != 0:
            self.result.is_failure("No persistent logs are saved in flash")
```

anta/tests/logging.py
```diff
@@ -0,0 +1,297 @@
...
+        * success: The test will pass if logs are generated.
+        * failure: The test will fail if logs are NOT generated.
+    """
+
+    name = "VerifyLoggingLogsGeneration"
+    description = "Verifies if logs are generated."
+    categories = ["logging"]
+    commands = [
+        AntaTestCommand(command="send log level informational message ANTA VerifyLoggingLogsGeneration validation"),
+        AntaTestCommand(command="show logging | tail", ofmt="text"),
```
gmuloc, line 151: could this trigger false positive if there are too many logs generated between two commands? (tail does not show ANTA log anymore ?)


Maybe something more like:

```
Leaf3#show logging last 30 s | grep ANTA
May 11 17:45:11 Leaf3 ConfigAgent: %SYS-6-LOGMSG_INFO: Message from arista on vty4 (10.95.70.233): ANTA VerifyLoggingLogsGeneration validation
```

anta/tests/logging.py
carl-baillargeon: No because the second you enable `logging persistent` logs are generated in `flash:/persist/messages` ;

```
NW-CORE(config)#more flash:/persist/messages
2023-05-11T11:19:19.425822-05:00 NW-CORE.jubacho.org rsyslogd: [origin software="rsyslogd" swVersion="8.2102.0" x-pid="22375" x-info="https://www.rsyslog.com"] exiting on signal 15.
2023-05-11T11:19:19.490017-05:00 NW-CORE.jubacho.org rsyslogd: imuxsock: Acquired UNIX socket '/run/systemd/journal/syslog' (fd 3) from systemd.  [v8.2102.0]
2023-05-11T11:19:19.490294-05:00 NW-CORE.jubacho.org rsyslogd: [origin software="rsyslogd" swVersion="8.2102.0" x-pid="22388" x-info="https://www.rsyslog.com"] start
```

So if the log file is actually 0 and `logging persistent` is enabled, it means there's a problem.

anta/tests/logging.py
carl-baillargeon: Agreed.

anta/tests/logging.py
gmuloc: let's put it once for now - maybe at the top of the file?

titom73 (approved):
LGTM!

PR #195, opened 2023-05-12
feat(anta.tests): Add field notices category

colinmacgiolla:
This PR creates a new test category - Field Notices, and moves the existing FN44 test to that.
It also creates a check for FN72, and adds the testcases


titom73 (approved):
Ship it !

PR #196, opened 2023-05-12
feat: Implement returning error code to anta functions

### LINKED ISSUE #192: Raise on error in the report part for CI

gmuloc:
Fixes #192  (even if apparently I can't even name my branches right!)

Behavior tested locally with the various new CLI flags

### DIFF
--- a/anta/cli/__init__.py
+++ b/anta/cli/__init__.py
@@ -55,11 +55,15 @@ from anta.result_manager.models import TestResult
     ),
     callback=setup_logging,
 )
+@click.option("--ignore-status", show_envvar=True, default=False, help="Always exit with success", is_eager=True)
+@click.option("--ignore-error", show_envvar=True, default=False, help="Only report failures and not errors", is_eager=True)
 def anta(ctx: click.Context, inventory: AntaInventory, **kwargs: Dict[str, Any]) -> None:
     # pylint: disable=unused-argument
     """Arista Network Test CLI"""
     ctx.ensure_object(dict)
     ctx.obj["inventory"] = inventory
+    ctx.obj["ignore_status"] = kwargs.get("ignore_status")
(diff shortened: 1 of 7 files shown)


anta/cli/__init__.py
```diff
@@ -55,11 +55,15 @@
     ),
     callback=setup_logging,
 )
+@click.option("--ignore-status", show_envvar=True, default=False, help="Always exit with success", is_eager=True)
+@click.option("--ignore-error", show_envvar=True, default=False, help="Only report failures and not errors", is_eager=True)
 def anta(ctx: click.Context, inventory: AntaInventory, **kwargs: Dict[str, Any]) -> None:
```
mtache, line 60: Possible to do that:
```suggestion
def anta(ctx: click.Context, inventory: AntaInventory, ignore_status: bool, ignore_error: bool, **kwargs: Dict[str, Any]) -> None:
```

anta/cli/__init__.py
```diff
@@ -55,11 +55,15 @@
     ),
     callback=setup_logging,
 )
+@click.option("--ignore-status", show_envvar=True, default=False, help="Always exit with success", is_eager=True)
```
mtache, line 58: is_eager is not needed here.
Also, defining options like this means you wants a boolean as parameter.
```
anta --help
  --ignore-status BOOLEAN         Always exit with success  [env var:
                                  ANTA_IGNORE_STATUS]
  --ignore-error BOOLEAN          Only report failures and not errors  [env
                                  var: ANTA_IGNORE_ERROR]
```
You need
```suggestion
@click.option("--ignore-status", is_flag=True, show_envvar=True, default=False, help="Always exit with success", is_eager=True)
```

Could you also update the insecure flag with the is_flag param=True param ?

mtache (changes requested):

mtache (approved):

PR #201, opened 2023-05-23
feat(anta.tests): Add new AAA test cases

carl-baillargeon:
Hello,

Please review this PR for new AAA test cases.

Thanks


anta/tests/aaa.py
```diff
@@ -0,0 +1,442 @@
...
+    """
+    if len(auth_types) > len(valid_auth_types):
+        raise ValueError(f"Too many parameters provided in auth_types. Valid parameters are: {valid_auth_types}")
+
+    for auth_type in auth_types:
+        if auth_type not in valid_auth_types:
+            raise ValueError(f"Wrong parameter provided in auth_types. Valid parameters are: {valid_auth_types}")
+
+
+# FIXME: AntaTestTemplate does not support 2 templates
```
titom73, line 42: Not sure if we should keep this todo in the code, or open an issue to track as it seems to be a major change required to support this feature.

titom73 (commented):
Looks very good. Just wondering if we need to keep comments as it is related to a major change (we could use issue to track it)

anta/tests/aaa.py
carl-baillargeon: Agreed let's open an issue!

titom73 (approved):
LGTM

PR #214, opened 2023-06-08
feat(anta.cli): added latest option to collect-tech-support

### LINKED ISSUE #198: collect tech-support files
Is it possible when using ANTA to collect the tech-support filles from a switch in the /mnt/flash/schedule/tech-support folder to only collect the latest file and not all the files?

mtache:
Fixes #198


carl-baillargeon (approved):
LGTM. Tested a few scenarios locally with 1-2-3 files.

titom73 (changes requested):
Nice option! Maybe we should update doc section to reflect this new option ?

titom73 (approved):
Ship It !

PR #218, opened 2023-06-16
fix(anta.tests): do not send 'show logging |' to EOS

### LINKED ISSUE #211: Bug: Complex commands from logging tests are sometimes failing
Intermittent errors when running some of the logging test cases that have complex commands.

```
ERROR    Command failed on NW-CORE: CLI command 2 of 2 'show logging | awk '/Trap logging/,/Sequence       models.py:239
                    numbers/{if(/Sequence numbers/) exit; print}'' failed: internal error
```

mtache:
Fixes #211

### DIFF
--- a/anta/tests/logging.py
+++ b/anta/tests/logging.py
@@ -7,11 +7,17 @@ from __future__ import annotations
 
 import logging
 import re
-from typing import Dict, List, Optional, cast
+from typing import List, Optional, cast
 
 from anta.models import AntaTest, AntaTestCommand
 
 logger = logging.getLogger(__name__)
+
+
+def _get_logging_states(command_output: str) -> str:
+    log_states = command_output.partition('\n\nExternal configuration:')[0]
+    logger.debug(f"Device logging states:\n{log_states}")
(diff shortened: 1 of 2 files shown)


anta/tests/logging.py
```diff
@@ -36,22 +42,20 @@ def test(self) -> None:
...
 
-        if "disabled" in log_output:
+        output = cast(str, self.instance_commands[0].output)
+
+        if "Persistent logging: disabled" in _get_logging_states(output):
             self.result.is_failure("Persistent logging is disabled")
             return
 
-        if not match or int(match.group(1)) == 0:
+        output = cast(str, self.instance_commands[0].output)
```
carl-baillargeon, line 53: We have to take the output of the second command here ; `self.instance_commands[1].output`

I think we should keep different output variable names like the initial code so it's not confusing.

mtache: Right, actually I made a mistake in this code.
Thanks !

carl-baillargeon (approved):
LGTM

Issue #206, opened 2023-06-07
Feat: Replace decorators for skipping tests with AntaTestFilter


carl-baillargeon:
Today we are using decorators on tests that should be skipped. We should leverage AntaTestFilter class instead for more flexibility.

Ex:

```
class VerifyBGPIPv4UnicastCount(AntaTest):
    test_filters = [SkipBGPFilter(family="ipv4")]

class SkipBGPFilter(AntaTestFilter):
    def __init__(self, family: str) -> None:
        """
        Class initialization with the BGP address family to verify.
        """
        self.family = family
```

carl-baillargeon:
#230

PR #230, opened 2023-06-27
feat(anta): Replace decorators for skipping tests with AntaTestFilter class

carl-baillargeon:
Hello,

WIP PR to integrate `AntaTestFilter` class to skip tests.

Related to #206


gmuloc:
Closing for now - we will revisit if we think it is worth it

PR #238, opened 2023-07-03
feat(anta.tests): Interface tests

carl-baillargeon:
Hello,

Please review and test this PR for new interface tests.

Thank you!


anta/tests/interfaces.py
```diff
@@ -321,3 +321,95 @@ def test(self) -> None:
...
+        if not mtu:
+            self.result.is_skipped(f"{self.__class__.name} did not run because mtu was not supplied")
+            return
+
+        command_output = cast(Dict[str, Dict[Any, Any]], self.instance_commands[0].output)
+
+        wrong_l3mtu_intf = []
+
+        for interface, values in command_output["interfaces"].items():
+            if not re.sub(r"\d+$", "", interface) in self.NOT_SUPPORTED_INTERFACES:
```
gmuloc, line 369: Think it would be better like this
```suggestion
            if re.sub(r"\d+$", "", interface) not in self.NOT_SUPPORTED_INTERFACES:
```

your version is correct but can be a bit more troublesome to understand: https://stackoverflow.com/questions/3481554/x-not-in-vs-not-x-in

carl-baillargeon: Yes I agree! Good catch thanks

gmuloc (approved):

PR #239, opened 2023-07-03
feat(anta.tests): New connectivity tests

carl-baillargeon:
Hello,

Please review and test this PR for new connectivity tests.

Thank you

### DIFF
--- a/anta/tests/connectivity.py
+++ b/anta/tests/connectivity.py
@@ -1,0 +1,49 @@
+"""
+Test functions related to various connectivity checks
+"""
+from __future__ import annotations
+
+import logging
+from typing import Any, Dict, List, Optional, cast
+
+from anta.models import AntaTest, AntaTestCommand, AntaTestTemplate
+
+logger = logging.getLogger(__name__)
+
+
+class VerifyReachability(AntaTest):
+    """
+    Test the network reachability to one or many destination IP(s).
+
+    Expected Results:
+        * success: The test will pass if all destination IP(s) are reachable.
+        * failure: The test will fail if one or many destination IP(s) are unreachable.
+        * error: The test will give an error if the destination IP(s) or the source interface/IP(s) are not provided as template_params.
+    """
+
+    name = "VerifyReachability"
+    description = "Test the network reachability to one or many destination IP(s)."
+    categories = ["connectivity"]
+    template = AntaTestTemplate(template="ping {dst} source {src} repeat 2")
+
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """
+        Run VerifyReachability validation.
+        """
+        
+        failures = []
+
+        for index, command in enumerate(self.instance_commands):
+            src, dst = (cast(Dict[str, str], command.template_params)["src"],
+                        cast(Dict[str, str], command.template_params)["dst"])
+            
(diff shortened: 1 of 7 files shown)


anta/tests/connectivity.py
```diff
@@ -0,0 +1,48 @@
...
+
+            command_output = cast(Dict[str, Dict[Any, Any]], self.instance_commands[index].output)
+            if "2 received" not in command_output["messages"][0]:
+                failures.append((src, dst))
+
+        if not failures:
+            self.result.is_success()
+
+        else:
+            self.result.is_failure(f"Connectivity test failed for the following source-destination pairs: {failures}")
```
gmuloc, line 48:
```suggestion
        self.result.is_success()
        
        for index, command in enumerate(self.instance_commands):
            src, dst = (cast(Dict[str, str], command.template_params)["src"], cast(Dict[str, str], command.template_params)["dst"])

            command_output = cast(Dict[str, Dict[Any, Any]], self.instance_commands[index].output)
            if "2 received" not in command_output["messages"][0]:
                failures.append((src, dst))

        if failures:
            self.result.is_failure(f"Connectivity test failed for the following source-destination pairs: {failures}")
```

anta/tests/connectivity.py
```diff
@@ -0,0 +1,48 @@
...
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """
+        Run VerifyReachability validation.
+        """
+
+        failures = []
+
+        for index, command in enumerate(self.instance_commands):
+            src, dst = (cast(Dict[str, str], command.template_params)["src"], cast(Dict[str, str], command.template_params)["dst"])
```
gmuloc, line 38: so the answer is most certainly yes but do you need casting for the template_params as well here?

anta/tests/connectivity.py
carl-baillargeon: Why should we set the result to success at the beginning, before verifying anything?

anta/tests/connectivity.py
carl-baillargeon: Yes :)

`anta/tests/connectivity.py:38: error: Value of type "Optional[Dict[str, str]]" is not indexable  [index]`

anta/tests/connectivity.py
gmuloc: Maybe this is better ? - this is only to remove comparison line.

```suggestion
        for index, command in enumerate(self.instance_commands):
            src, dst = (cast(Dict[str, str], command.template_params)["src"], cast(Dict[str, str], command.template_params)["dst"])

            command_output = cast(Dict[str, Dict[Any, Any]], self.instance_commands[index].output)
            if "2 received" not in command_output["messages"][0]:
                failures.append((src, dst))

        self.result.is_success()
        if failures:
            self.result.is_failure(f"Connectivity test failed for the following source-destination pairs: {failures}")
```

(though having it above the for loop is more or less the same except if one of the lines in the loop raises an Exception but it will be caught by the decorator so this is the same - you can see this as "by default the test succeeds except if there is a failure")

anta/tests/connectivity.py
gmuloc: ok that's something it would be nice to get rid of :)

gmuloc (approved):

PR #246, opened 2023-07-04
feat(anta.tests): MLAG tests update

carl-baillargeon:
Hello,

Please review and test these updated/new MLAG tests.

Thank you

### DIFF
--- a/anta/tests/mlag.py
+++ b/anta/tests/mlag.py
@@ -94,6 +94,7 @@ class VerifyMlagConfigSanity(AntaTest):
         * success: The test will pass if there are NO MLAG config-sanity inconsistencies.
         * failure: The test will fail if there are MLAG config-sanity inconsistencies.
         * skipped: The test will be skipped if MLAG is 'disabled'.
+        * error: The test will give an error if 'mlagActive' is not found in the JSON response.
     """
 
     name = "VerifyMlagConfigSanity"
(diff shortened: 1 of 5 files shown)


anta/tests/mlag.py
```diff
@@ -1,93 +1,233 @@
...
+        """
+        Run VerifyMlagConfigSanity validation
+        """
 
         command_output = cast(Dict[str, Dict[str, Any]], self.instance_commands[0].output)
 
-        if "mlagActive" not in command_output.keys():
-            self.result.is_error("Incorrect JSON response - mlagActive state not found")
-        elif command_output["mlagActive"] is False:
+        if command_output["mlagActive"] is False:
```
gmuloc, line 112: Why is this gone:

```
if "mlagActive" not in command_output.keys():
            self.result.is_error("Incorrect JSON response - mlagActive state not found")
```

I guess it was certainly there for some corner case we may not be able to repro today but we shoudl keep it

anta/tests/mlag.py
```diff
@@ -1,93 +1,233 @@
...
             self.result.is_skipped("MLAG is disabled")
-        elif len(command_output["globalConfiguration"]) > 0 or len(command_output["interfaceConfiguration"]) > 0:
-            self.result.is_failure()
-            if len(command_output["globalConfiguration"]) > 0:
-                self.result.is_failure("MLAG config-sanity returned Global inconsistancies: " f"{command_output['globalConfiguration']}")
-            if len(command_output["interfaceConfiguration"]) > 0:
-                self.result.is_failure("MLAG config-sanity returned Interface inconsistancies: " f"{command_output['interfaceConfiguration']}")
+            return
```
gmuloc, line 114: why return here and using `elif` afterwards ?

anta/tests/mlag.py
```diff
@@ -1,93 +1,233 @@
...
+
+        if len(command_output["globalConfiguration"]) == 0 and len(command_output["interfaceConfiguration"]) == 0:
+            self.result.is_success()
+
+        elif len(command_output["globalConfiguration"]) > 0:
+            self.result.is_failure(f"MLAG config-sanity returned Global inconsistancies: {command_output['globalConfiguration']}")
+
+        elif len(command_output["interfaceConfiguration"]) > 0:
+            self.result.is_failure(f"MLAG config-sanity returned Interface inconsistancies: {command_output['interfaceConfiguration']}")
```
gmuloc, line 123: This is not the same as was in the test originallly

Here you cannot catch both interfce and global inconsistencies so it should be changed back to what it was

gmuloc, line 123: we should probably add sommething if we reach the last elif otherwise (or is it actually possible ? I think not so maybe we should rewrite it in another way)

anta/tests/mlag.py
```diff
@@ -1,93 +1,233 @@
...
+        if detection_delay is None or errdisabled is None or recovery_delay is None or recovery_delay_non_mlag is None:
+            self.result.is_skipped(
+                f"{self.__class__.name} did not run because detection_delay, errdisabled, recovery_delay or recovery_delay_non_mlag were not supplied"
+            )
+            return
+
+        if errdisabled:
+            action = "errdisableAllInterfaces"
         else:
+            action = "none"
```
gmuloc, line 209:
```suggestion
        errdisabled_action = "errdisableAllInterfaces" if errdisabled else "none"
```

gmuloc (changes requested):

anta/tests/mlag.py
```diff
@@ -1,93 +1,234 @@
...
-                self.result.is_failure("MLAG config-sanity returned Interface inconsistancies: " f"{command_output['interfaceConfiguration']}")
+            return
+
+        keys_to_verify = ["globalConfiguration", "interfaceConfiguration"]
+        verified_output = {key: get_value(command_output, key) for key in keys_to_verify}
+
+        if not any(verified_output.values()):
+            self.result.is_success()
         else:
+            self.result.is_failure(f"MLAG config-sanity returned inconsistancies: {verified_output}")
```
gmuloc, line 127:
```suggestion
            self.result.is_failure(f"MLAG config-sanity returned inconsistencies: {verified_output}")
```

PR #256, opened 2023-07-07
fix(anta.cli): Add --group-by in nrfu table command

titom73:
Add `--group-by` option removed in 0.6.0 to provides summary view based on
a per host or per test approach.

### DIFF
--- a/anta/cli/nrfu/commands.py
+++ b/anta/cli/nrfu/commands.py
@@ -12,6 +12,7 @@ from typing import List, Optional
 
 import click
 
+from anta.cli.console import console
 from anta.cli.utils import parse_tags, return_code
 from anta.models import AntaTest
 from anta.result_manager import ResultManager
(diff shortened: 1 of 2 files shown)


anta/cli/nrfu/commands.py
```diff
@@ -27,13 +28,19 @@
 @click.option("--tags", "-t", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags)
 @click.option("--device", "-d", help="Show a summary for this device", type=str, required=False)
 @click.option("--test", "-t", help="Show a summary for this test", type=str, required=False)
-def table(ctx: click.Context, tags: Optional[List[str]], device: Optional[str], test: Optional[str]) -> None:
+@click.option(
+    "--group-by", default=None, type=click.Choice(["none", "host", "test"], case_sensitive=False), help="Group result by test or host. default none", required=False
+)
+def table(ctx: click.Context, tags: Optional[List[str]], device: Optional[str], test: Optional[str], group_by: str) -> None:
     """ANTA command to check network states with table result"""
+    console.print("\n")
```
mtache, line 36: I was wondering where these newlines came from in my terminal :)
When using DEBUG mode, this line is irrelevant (at least from its original purpose).
Could you remove this and/or add it to the `print_settings()`/`print_table()` functions ?

anta/cli/nrfu/commands.py
```diff
@@ -27,13 +28,19 @@
 @click.option("--tags", "-t", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags)
 @click.option("--device", "-d", help="Show a summary for this device", type=str, required=False)
 @click.option("--test", "-t", help="Show a summary for this test", type=str, required=False)
-def table(ctx: click.Context, tags: Optional[List[str]], device: Optional[str], test: Optional[str]) -> None:
+@click.option(
+    "--group-by", default=None, type=click.Choice(["none", "host", "test"], case_sensitive=False), help="Group result by test or host. default none", required=False
```
mtache, line 32: The `none` value is not used, I think we can just remove it.
Could we change the `host` value to `device` which makes it coherent with the other options ?

anta/cli/nrfu/utils.py
```diff
@@ -34,15 +34,19 @@ def print_settings(context: click.Context, report_template: Optional[pathlib.Pat
...
     """Print result in a table"""
     reporter = ReportTable()
-    if device:
-        console.print(reporter.report_summary_hosts(result_manager=results, host=device))
-    if test:
-        console.print(reporter.report_summary_tests(result_manager=results, testcase=test))
-    if device is None and test is None:
+    if device or test:
+        message = "The test and device options are not supported yet."
+        console.print(Panel.fit(message, style="red", title="[red]Warning"))
```
mtache, line 42: The lines that implements these options are being deleted by this PR, let's keep these lines that implements the simple use case: filter one device and/or one test.

mtache (changes requested):

anta/cli/nrfu/commands.py
titom73: Changed

PR #276, opened 2023-07-18
bump: Relax PyYAML dependency due to https://github.com/yaml/pyyaml/issues/601

u1735067:


mtache (approved):
Thanks for catching this and the PR !

PR #287, opened 2023-07-24
feat(anta.tests)!: Update VXLAN tests

carl-baillargeon:
* rename `VerifyVxlan` to `VerifyVxlanInterface`
* update documentation

### DIFF
--- a/anta/tests/vxlan.py
+++ b/anta/tests/vxlan.py
@@ -9,6 +9,9 @@ class VerifyVxlan1Interface(AntaTest):
     """
     This test verifies if the Vxlan1 interface is configured and 'up/up'.
 
+    !!! warning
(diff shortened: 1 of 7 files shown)


anta/tests/vxlan.py
```diff
@@ -5,19 +5,26 @@
 from anta.models import AntaCommand, AntaTest
 
 
-class VerifyVxlan(AntaTest):
+class VerifyVxlan1Interface(AntaTest):
```
gmuloc, line 8: if we change the name we probably need to write some kind of note somewhere ... in the docs ?

carl-baillargeon: Agreed! Done.

gmuloc (approved):

PR #288, opened 2023-07-24
fix(anta.cli): Add --group-by in nrfu table command

titom73:
Add `--group-by` option removed in `0.6.0` to provides summary view based on
a per host or per test approach.

Options `-d/-t` now list all entries related to either device or test-case

Clean PR for old #256


gmuloc (approved):
Added the doc myself and had @titom73 review

so we are good to go :)

Issue #269, opened 2023-07-12
feat: TestResults should have test input parameters


gmuloc:
e.g.

if I run

```
anta.tests.connectivity:
  - VerifyReachability: 
     template_params:
       - src: 1.1.1.1
       - dst: 10.0.0.01
  - VerifyReachability: 
     template_params:
       - src: 1.1.1.1
       - dst: 10.0.0.02
```

Then there is no way to distinguish one test from another in the test results

PR #292, opened 2023-07-25
feat: Add input parameters to TestResult

carl-baillargeon:
PR to add test input parameters + template params to TestResult object.

Fixes: #269


gmuloc:
Putting in draft state in favor of #296  for now

PR #293, opened 2023-07-25
doc: Generate svg from rich console output

gmuloc:
Target- generate the screenshot in documentation using this


gmuloc:
Usage

```
python docs/scripts/generate_svg.py anta <input_param>
```

PR #337, opened 2023-08-04
ci: Add comment if docs/ is not updated

titom73:
Add a step to comment PR if documentation folder is not updated

image


mtache (approved):
🐜 🪖

PR #340, opened 2023-08-04
feat: prompt passwords from stdin if not provided via CLI options

mtache:
Use click [password_option](https://click.palletsprojects.com/en/8.1.x/api/#click.password_option) and [prompt](https://click.palletsprojects.com/en/8.1.x/api/#click.prompt) to prompt EOS passwords if not provided via environment variables or CLI options.
This is to provide a more secure way to provide passwords without using environment variables.

### DIFF
--- a/anta/cli/__init__.py
+++ b/anta/cli/__init__.py
@@ -15,7 +15,7 @@ from anta.cli.debug import commands as debug_commands
 from anta.cli.exec import commands as exec_commands
 from anta.cli.get import commands as get_commands
 from anta.cli.nrfu import commands as check_commands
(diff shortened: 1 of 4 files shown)


titom73 (approved):
Looks good

```bash
❯ anta nrfu --catalog .personal/catalog-class.yml table
Please enter a password to connect to EOS: 
Repeat for confirmation: 
...
```

But password is asked even for `--help`

```bash
❯ anta nrfu --help
Please enter a password to connect to EOS: 
Aborted!
```

titom73 (approved):
Looks good to me! 🐜

PR #344, opened 2023-08-07
feat: add --log-file option to send logs to a file

mtache:
- test: remove generated inventory files
- Refactored return_code into exit_with_code
- feat: add --log-file option to send logs to a file

### DIFF
--- a/anta/__init__.py
+++ b/anta/__init__.py
@@ -15,7 +15,7 @@ __credits__ = [
 ]
 __copyright__ = "Copyright 2022, Arista EMEA AS"
 
-# Global ANTA debug environment variable. Can be set using the cli 'anta --debug'.
+# Global ANTA debug mode environment variable
 __DEBUG__ = bool(os.environ.get("ANTA_DEBUG", "").lower() == "true")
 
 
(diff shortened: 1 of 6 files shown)


docs/README.md
```diff
@@ -65,8 +71,8 @@ Commands:
   nrfu   Run NRFU against inventory devices
 ```
 
-!!! info
-    `username`, `password`, `enable`, and `enable-password` values are the same for all devices
+> [!NOTE]
```
titom73, line 74: Seems GH markdown flavor is not supported by material for mkdocs. Rendering is not optimal

```
[!NOTE] username, password, enable-password, enable, timeout and insecure values are the same for all devices
```

mtache: Yeah, the opposite is also true.. I chose the GitHub-compatible `[!NOTE]` annotation for the GitHub README page because `!!! info` is not rendered.
Maybe should we just use:
```
> username, password, enable-password, enable, timeout and insecure values are the same for all devices
```
So GitHub and mkdocs are happy ?

titom73: Yes, let's do that ☮️

PR #349, opened 2023-08-08
doc: Implement UML diagram for classes

titom73:
# Overview

Implementation of visual representation of object with their dependencies.

Pylint comes with a tool named `pyreverse` that can generate UML diagram for
Python class and modules.

# Usage

```
pyreverse anta --colorized -a1 -s1 -o jpeg -m true -k --output-directory docs/imgs/uml/ -c anta.device.AntaDevice
```

# Requirements

To be able to generate JPEG output, it is important to install [graphviz](https://graphviz.org/)


mtache:
Could we have this beautiful line of shell in the contribution guide ? :)

mtache (approved):
Awesome diagrams !

Issue #336, opened 2023-08-03
False positive on VerifyCoredump test


carl-baillargeon:
The presence of minidump empty folder in /var/core is causing `VerifyCoredump` to fail:

`Core-dump(s) have been found: minidump`

Maybe the test should verify if there are files starting with `core.*`

https://arista.my.site.com/AristaCommunity/s/article/troubleshooting-filesystem-full-issues

PR #350, opened 2023-08-08
fix(anta.tests): Update VerifyCoredump to JSON output and skip minidump folder

carl-baillargeon:
Update VerifyCoredump to JSON output and skip minidump folder.

Fixes: #336


mtache (approved):
Work on my machine 🐜

Issue #334, opened 2023-08-03
Better formatting of test result messages


carl-baillargeon:
Some tests with template_params as input could have better result messages for improved readability in the reports.
```
STP instance for VLAN 10 is not configured 
STP instance for VLAN 20 is not configured
```
It could be something like:
```
STP instance is not configured for the following VLAN(s): 10, 20
```

PR #351, opened 2023-08-08
fix(anta.tests): Update STP AntaTemplate tests

carl-baillargeon:
Update STP AntaTemplate tests for better results output.

Fixes: #334


titom73 (approved):
LGTM!

Issue #332, opened 2023-08-03
VerifySyslog test is way too CPU hungry


carl-baillargeon:
`VerifySyslog` test takes a lot of time to complete (close to 1 minute) and is very CPU demanding when the log buffer is large.

CPU usage over 75% on 7280SR3K-48YC8.

A period of 7 days might be too long.

Also, the test should be moved from the `system` module to the `logging` module.

Issue #342, opened 2023-08-04
Implementing caching for `AntaDevice.collect()`


mtache:
Investigate caching implementing with https://github.com/aio-libs/aiocache

PR #356, opened 2023-08-09
fix(anta.tests)!: Update VerifySyslog

carl-baillargeon:
Move anta.tests.system.VerifySyslog test to anta.tests.logging.VerifyLoggingErrors.

For now, the test has been modified to verify syslog messages with a severity of ERRORS or higher instead of WARNING.

The last 7 days has been removed as well to optimize test performance on devices with large log buffers.

This test can be revisited in the future once we implement caching in #342.

Fixes: #332


titom73 (approved):
LGTM

PR #377, opened 2023-09-06
feat(anta): Add tag mapping between devices and tests

titom73:
# Overview

The `--tags` option can be used to target specific devices in your inventory and run only tests configured with this specific tags from your catalog. Expected behaviour is provided below:

| Command | Description |
| ------- | ----------- |
| `none` | Run all tests on all devices according `tag` definition in your inventory and tests catalog. And tests with no tag are executed on all devices|
| `--tags leaf` | Run all tests marked with `leaf` tag on all devices configured with `leaf` tag.<br/> All other tags are ignored |
| `--tags leaf,spine` | Run all tests marked with `leaf` tag on all devices configured with `leaf` tag.<br/>Run all tests marked with `spine` tag on all devices configured with `spine` tag.<br/> All other tags are ignored |

> **warning**
> This PR is not implementing device filtering

#### Inventory

```yaml
---
anta_inventory:
  hosts:
  - host: 192.168.0.10
    name: spine01
    tags: ['fabric', 'spine']
  - host: 192.168.0.11
    name: spine02
    tags: ['fabric', 'spine']
  - host: 192.168.0.12
    name: leaf01
    tags: ['fabric', 'leaf']
  - host: 192.168.0.13
    name: leaf02
    tags: ['fabric', 'leaf']
  - host: 192.168.0.14
    name: leaf03
    tags: ['fabric', 'leaf']
  - host: 192.168.0.15
    name: leaf04
    tags: ['fabric', 'leaf']
```

#### Tests

```yaml

anta.tests.system:
  - VerifyUptime:
      minimum: 10
      filters:
        tags: ['fabric']
  - VerifyReloadCause:
      filters:
        tags: ['leaf', spine']
  - VerifyCoredump:
  - VerifyAgentLogs:
  - VerifyCPUUtilization:
      filters:
        tags: ['leaf']
  - VerifyMemoryUtilization:
  - VerifyFileSystemUtilization:
  - VerifyNTP:

anta.tests.mlag:
  - VerifyMlagStatus:

anta.tests.interfaces:
  - VerifyL3MTU:
      mtu: 1500
      filters:
        tags: ['demo']
```

### CLI output

image

### DIFF
--- a/anta/cli/nrfu/commands.py
+++ b/anta/cli/nrfu/commands.py
@@ -23,7 +23,9 @@ logger = logging.getLogger(__name__)
 
 @click.command()
 @click.pass_context
-@click.option("--tags", default="all", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags)
+@click.option(
+    "--tags", default="all", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags, show_default=True
+)
 @click.option("--device", "-d", help="Show a summary for this device", type=str, required=False)
 @click.option("--test", "-t", help="Show a summary for this test", type=str, required=False)
 @click.option(
@@ -40,7 +42,9 @@ def table(ctx: click.Context, tags: list[str], device: str | None, test: str | N
 
 @click.command()
 @click.pass_context
-@click.option("--tags", "-t", default="all", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags)
+@click.option(
+    "--tags", "-t", default="all", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags, show_default=True
+)
 @click.option(
     "--output",
     "-o",
@@ -62,7 +66,9 @@ def json(ctx: click.Context, tags: list[str], output: pathlib.Path | None) -> No
 
 @click.command()
 @click.pass_context
-@click.option("--tags", "-t", default="all", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags)
+@click.option(
+    "--tags", "-t", default="all", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags, show_default=True
+)
 @click.option("--search", "-s", help="Regular expression to search in both name and test", type=str, required=False)
 @click.option("--skip-error", help="Hide tests in errors due to connectivity issue", default=False, is_flag=True, show_default=True, required=False)
 def text(ctx: click.Context, tags: list[str], search: str | None, skip_error: bool) -> None:
@@ -92,7 +98,9 @@ def text(ctx: click.Context, tags: list[str], search: str | None, skip_error: bo
     required=False,
     help="Path to save report as a file",
 )
-@click.option("--tags", "-t", default="all", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags)
+@click.option(
+    "--tags", "-t", default="all", help="List of tags using comma as separator: tag1,tag2,tag3", type=str, required=False, callback=parse_tags, show_default=True
+)
 def tpl_report(ctx: click.Context, tags: list[str], template: pathlib.Path, output: pathlib.Path | None) -> None:
     """ANTA command to check network state with templated report"""
     print_settings(ctx, template, output)
--- a/anta/cli/nrfu/commands.py
+++ b/anta/cli/nrfu/commands.py
@@ -32,11 +32,11 @@ logger = logging.getLogger(__name__)
 @click.option(
     "--group-by", default=None, type=click.Choice(["device", "test"], case_sensitive=False), help="Group result by test or host. default none", required=False
 )
-def table(ctx: click.Context, tags: str, device: Optional[str], test: Optional[str], group_by: str) -> None:
+def table(ctx: click.Context, tags: list[str], device: Optional[str], test: Optional[str], group_by: str) -> None:
     """ANTA command to check network states with table result"""
     print_settings(ctx)
     with anta_progress_bar() as AntaTest.progress:
-        asyncio.run(main(ctx.obj["result_manager"], ctx.obj["inventory"], ctx.obj["catalog"], tags=tags.split(",")))
+        asyncio.run(main(ctx.obj["result_manager"], ctx.obj["inventory"], ctx.obj["catalog"], tags_cli=tags))
     print_table(results=ctx.obj["result_manager"], device=device, group_by=group_by, test=test)
     exit_with_code(ctx)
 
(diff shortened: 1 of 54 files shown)


docs/usage-inventory-catalog.md
```diff
@@ -104,6 +104,20 @@ anta.tests.configuration:
   - VerifyRunningConfigDiffs:
 ```
 
+All tests can be configure with a list of user's defined tags. These tags will be mapped with devices tags when cli is used with `--tags` option. If test is configure without a list of tags, the default one will be used: `['all']`
```
gmuloc, line 107:
```suggestion
All tests can be configured with a list of user defined tags. These tags will be mapped with device tags when `nrfu` cli is called with the `--tags` option. If a test is configured in the catalog without a list of tags, the default one will be used: `['all']`
```

anta/runner.py
```diff
@@ -49,14 +49,18 @@ async def main(
...
-        test_inputs = test[1]
-        try:
-            # Instantiate AntaTest object
-            test_instance = test_class(device=device, inputs=test_inputs)
-            coros.append(test_instance.test(eos_data=None))
-        except Exception as e:  # pylint: disable=broad-exception-caught
-            message = "Error when creating ANTA tests"
-            anta_log_exception(e, message, logger)
+        test_inputs = test[1] or {}
+        test_tags = test_inputs.get("tags") if "tags" in test_inputs.keys() else ["all"]
```
gmuloc, line 53:
```suggestion
        test_tags = test_inputs.get("tags", ["all"])
```

anta/runner.py
```diff
@@ -49,14 +49,18 @@ async def main(
...
         test_class = test[0]
-        test_inputs = test[1]
-        try:
-            # Instantiate AntaTest object
-            test_instance = test_class(device=device, inputs=test_inputs)
-            coros.append(test_instance.test(eos_data=None))
-        except Exception as e:  # pylint: disable=broad-exception-caught
-            message = "Error when creating ANTA tests"
-            anta_log_exception(e, message, logger)
+        test_inputs = test[1] or {}
```
gmuloc, line 52: not related to this line but should we rename `tags` `device_tags` for clarity?

anta/runner.py
```diff
@@ -49,14 +49,18 @@ async def main(
...
+        test_tags = test_inputs.get("tags") if "tags" in test_inputs.keys() else ["all"]
+        if tags is None:
+            tags = ["all"]
+        if any(t in test_tags for t in tags):  # type: ignore
```
gmuloc, line 56: can we remove the type ignore?

anta/runner.py
gmuloc: Do we need this something more for this kind of scenario:
```
anta.tests.system:
  - VerifyUptime:
      minimum: 10
      tags: null
```
?

anta/runner.py
```diff
@@ -49,14 +49,18 @@ async def main(
...
-            # Instantiate AntaTest object
-            test_instance = test_class(device=device, inputs=test_inputs)
-            coros.append(test_instance.test(eos_data=None))
-        except Exception as e:  # pylint: disable=broad-exception-caught
-            message = "Error when creating ANTA tests"
-            anta_log_exception(e, message, logger)
+        test_inputs = test[1] or {}
+        test_tags = test_inputs.get("tags") if "tags" in test_inputs.keys() else ["all"]
+        if tags is None:
+            tags = ["all"]
```
gmuloc, line 55: Can we not use `device.tags` here? we make sure that "all" is always in there

anta/models.py
```diff
@@ -244,6 +245,7 @@ class Input(BaseModel):
         """
 
         model_config = ConfigDict(extra="forbid")
+        tags: Optional[List[str]] = ["all"]
```
gmuloc, line 248: wondering if we should not prefix this in case someone would need `tags` in a test one day?

anta/cli/nrfu/commands.py
```diff
@@ -68,6 +72,8 @@ def json(ctx: click.Context, tags: Optional[List[str]], output: Optional[pathlib
 def text(ctx: click.Context, tags: Optional[List[str]], search: Optional[str], skip_error: bool) -> None:
     """ANTA command to check network states with text result"""
     print_settings(ctx)
+    if tags is None:
+        tags = ["all"]
     with anta_progress_bar() as AntaTest.progress:
         asyncio.run(main(ctx.obj["result_manager"], ctx.obj["inventory"], ctx.obj["catalog"], tags=tags))
```
gmuloc, line 78:
```suggestion
    with anta_progress_bar() as AntaTest.progress:
        asyncio.run(main(ctx.obj["result_manager"], ctx.obj["inventory"], ctx.obj["catalog"], tags=tags or ["all"]))
```

gmuloc (commented):
I think we probably need to add in the doc some example of "using tags" and saying what test will be executed or not

anta/runner.py
```diff
@@ -49,14 +49,18 @@ async def main(
...
+        if any(t in test_tags for t in tags):  # type: ignore
+            try:
+                # Instantiate AntaTest object
+                test_instance = test_class(device=device, inputs=test_inputs)
+                coros.append(test_instance.test(eos_data=None))
+            except Exception as e:  # pylint: disable=broad-exception-caught
+                message = "Error when creating ANTA tests"
+                anta_log_exception(e, message, logger)
```
mtache, line 63: I think this logic should not be in the module `anta.runner`.
Instead, it should be somewhere in the `AntaTest` constructor, after the input validation (you want to validate the inputs which contains the tags).
It will be a bit different in the report since you will skip the test. But IMHO it makes more sense (you can even write a clear message on why the test has been skipped with the tags values of the device and the test).

mtache (changes requested):
What do you think about having the logic below?
Init:
- Assuming tag: `leaf` is applied to some devices and tests.
- All tests are marked with tag `all` by default
- All devices are not marked with any tag by default
Logic:
- Tests with tag `all` are run on every devices
- A test with tag `all` cannot have any other tag (`all` includes all the tags)
- All the tests marked with tag: `leaf` are executed on all devices marked with tag: `leaf`
- All tests that do not have any tags are skipped
- All tests marked with a tag that do not match any device are skipped

anta/cli/nrfu/commands.py
gmuloc: actually @mtache proposal to use click default is better than this :)

anta/runner.py
gmuloc: could end up with a loooot of skipped result for this reason - probably better to either add a some result state or to never add the test to the ResultManager

titom73: Exactly. Also, runner is in charge of loading tests and execute them, so it seems better to not load tests we don't want to run with anta

anta/models.py
```diff
@@ -219,6 +219,7 @@ def test(self) -> None:
     description: ClassVar[str]
     categories: ClassVar[list[str]]
     commands: ClassVar[list[Union[AntaTemplate, AntaCommand]]]
+    tags: ClassVar[list[str]]
```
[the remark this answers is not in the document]
titom73: Fixed

anta/cli/nrfu/commands.py
```diff
@@ -34,6 +34,8 @@
 def table(ctx: click.Context, tags: Optional[List[str]], device: Optional[str], test: Optional[str], group_by: str) -> None:
     """ANTA command to check network states with table result"""
     print_settings(ctx)
+    if tags is None:
```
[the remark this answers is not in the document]
titom73: Fixed

anta/runner.py
titom73: It can be remove after code sanity

titom73: Fixed

titom73:
> I think we probably need to add in the doc some example of "using tags" and saying what test will be executed or not

I have added a dedicated page for tags and behaviours

mtache (approved):
Ship it 🐜

PR #394, opened 2023-09-14
feat(anta.tests)!: Implement caching

### LINKED ISSUE #342: Implementing caching for `AntaDevice.collect()`

carl-baillargeon:
# Description

Implement caching with `aiocache` library. 

Cache key design:

`<device_name>:<uid>`

The `uid` is a new property attribute  of `AntaCommand`, which is a UID generated from the command, version, revision and output format.

Caching is implemented in the `AntaDevice.collect_commands` async function where the inner `collect_command` function first checks if the `command.uid` is present in the cache and if it is, `command.output` will be populate by the cache. If not, the `collect()` function will be run as usual and the output will be saved in the cache.

Fixes #342 

# Tasks:
- [ ] Add a knob in AntaTest to activate/deactivate caching for a specific test
- [ ] Add logging with `aiocache` plugins for cache hit ratio, etc.
- [ ] Implement `click` CLI options to tweak caching (TTL, etc.)
- [ ] Maybe need to add a coroutine to check and clean-up unused Locks.
- [ ] Add functionalities to support external backend caches like Redis/memcached - future PR maybe.
- [ ] Documentation + examples

Caveats:
- Some tests in ANTA will break with caching. Logging tests are a good example where we generate a log message and check for that log right after. That's why we need a way to activate/deactivate caching for specific tests.
- Since there is a lot of varied UIDs and each device has its own instance of the `AntaDevice` class, we might need a lock cleanup strategy to optimize performance when running thousands of tests.

# Checklist:

### DIFF
--- a/anta/device.py
+++ b/anta/device.py
@@ -11,6 +11,8 @@ from pathlib import Path
 from typing import Any, Dict, Iterator, List, Literal, Optional, Tuple, Union
 
 import asyncssh
+from aiocache import Cache
+from aiocache.serializers import PickleSerializer
 from aioeapi import Device, EapiCommandError
 from asyncssh import SSHClientConnection, SSHClientConnectionOptions
 from httpx import ConnectError, HTTPError
@@ -49,6 +51,7 @@ class AntaDevice(ABC):
         self.tags: List[str] = tags if tags is not None else []
         self.is_online: bool = False
         self.established: bool = False
+        self.cache = Cache(cache_class=Cache.REDIS, ttl=60, serializer=PickleSerializer(), namespace=self.name, endpoint="10.22.10.6", password="secret")
 
         # Ensure tag 'all' is always set
         if DEFAULT_TAG not in self.tags:
@@ -96,7 +99,21 @@ class AntaDevice(ABC):
         Args:
             commands: the commands to collect
         """
-        await asyncio.gather(*(self.collect(command=command) for command in commands))
+        non_cached_commands = []
+        for command in commands:
+            cached_output = await self.cache.get(command.uid)
+            if cached_output is not None:
+                print("Cache hit!")
+                command.output = cached_output
+            else:
+                print("Cache miss")
+                non_cached_commands.append(command)
+
+        await asyncio.gather(*(self.collect(command=command) for command in non_cached_commands))
+
+        for command in non_cached_commands:
+            print(f"Adding command {command.uid} to cache")
+            await self.cache.set(command.uid, command.output, ttl=60)
 
     @abstractmethod
     async def refresh(self) -> None:
--- a/anta/device.py
+++ b/anta/device.py
@@ -7,12 +7,12 @@ ANTA Device Abstraction Module
 import asyncio
 import logging
 from abc import ABC, abstractmethod
+from collections import defaultdict
 from pathlib import Path
-from typing import Any, Dict, Iterator, List, Literal, Optional, Tuple, Union
+from typing import Any, DefaultDict, Dict, Iterator, List, Literal, Optional, Tuple, Union
 
 import asyncssh
 from aiocache import Cache
-from aiocache.serializers import PickleSerializer
 from aioeapi import Device, EapiCommandError
 from asyncssh import SSHClientConnection, SSHClientConnectionOptions
 from httpx import ConnectError, HTTPError
@@ -51,8 +51,8 @@ class AntaDevice(ABC):
         self.tags: List[str] = tags if tags is not None else []
         self.is_online: bool = False
         self.established: bool = False
-        self.cache = Cache(cache_class=Cache.REDIS, ttl=60, serializer=PickleSerializer(), namespace=self.name, endpoint="10.22.10.6", password="secret")
-        self.cache_lock = asyncio.Lock()
+        self.cache: Cache = Cache(cache_class=Cache.MEMORY, ttl=60, namespace=self.name)
+        self.cache_locks: DefaultDict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
 
         # Ensure tag 'all' is always set
         if DEFAULT_TAG not in self.tags:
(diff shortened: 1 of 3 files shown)


anta/device.py
```diff
@@ -96,7 +100,19 @@ async def collect_commands(self, commands: List[AntaCommand]) -> None:
         Args:
             commands: the commands to collect
         """
-        await asyncio.gather(*(self.collect(command=command) for command in commands))
+
+        async def collect_command(command: AntaCommand) -> None:
+            async with self.cache_locks[command.uid]:
+                cached_output = await self.cache.get(command.uid)  # pylint: disable=no-member
+                if cached_output is not None:
+                    print(f"Cache hit for {command.command} on {self.name}")
```
mtache, line 108: Use `logger`

anta/device.py
```diff
@@ -96,7 +100,19 @@ async def collect_commands(self, commands: List[AntaCommand]) -> None:
         Args:
             commands: the commands to collect
         """
-        await asyncio.gather(*(self.collect(command=command) for command in commands))
+
+        async def collect_command(command: AntaCommand) -> None:
```
mtache, line 104: I would implement `collect()` in `AntaDevice` with the caching logic. This logic should call a private method `_collect()` that would be an abstract method implemented in `AntaDevice` subclasses.

anta/models.py
```diff
@@ -107,9 +108,11 @@ class AntaCommand(BaseModel):
         version: eAPI version - valid values are 1 or "latest" - default is "latest"
         revision: eAPI revision of the command. Valid values are 1 to 99. Revision has precedence over version.
         ofmt: eAPI output - json or text - default is json
+        output: Output of the command populated by the collect() function
         template: AntaTemplate object used to render this command
-        params: dictionary of variables with string values to render the template
+        params: Dictionary of variables with string values to render the template
         failed: If the command execution fails, the Exception object is stored in this field
+        cache: Enable or disable caching for this AntaCommand
```
gmuloc, line 115: Add that it is True by default

gmuloc: and maybe let's rename the parameter `use_cache` ?

anta/device.py
```diff
@@ -52,11 +55,21 @@ def __init__(self, name: str, tags: Optional[list[str]] = None) -> None:
         self.tags: list[str] = tags if tags is not None else []
         self.is_online: bool = False
         self.established: bool = False
+        self.cache: Cache = Cache(cache_class=Cache.MEMORY, ttl=60, namespace=self.name, plugins=[HitMissRatioPlugin()])
```
gmuloc, line 58: Need to be added to the doc string

anta/device.py
```diff
@@ -75,23 +88,51 @@ def __eq__(self, other: object) -> bool:
...
+
+        Effects:
+            - Updates `command.output` with the cached or collected data.
+            - Logs when data is retrieved from cache.
+        """
+        async with self.cache_locks[command.uid]:
+            cached_output = None
+
+            if command.cache:
+                cached_output = await self.cache.get(command.uid)  # pylint: disable=no-member
```
gmuloc, line 127: Can we get read of pylint disable?

gmuloc (commented):
- [x] Let's add a way in the inventory to disable the `cache` per device if needed
- [ ] Let's add a CLI to disable the cache completely as well (as mentioned in your PR)
- [ ] Documentation to add as well with examples maybe
    * Disabling cache for a command
    * Disabling cache globally
    * Disabling cache for a device from the inventory

docs/README.md
```diff
@@ -32,52 +32,8 @@ $ pip install anta
...
-  --ignore-error                  Only report failures and not errors  [env
-                                  var: ANTA_IGNORE_ERROR]
-  --help                          Show this message and exit.
-
-Commands:
-  debug  Debug commands for building ANTA
-  exec   Execute commands to inventory devices
-  get    Get data from/to ANTA
-  nrfu   Run NRFU against inventory devices
+--8<-- "anta_help.txt"
```
gmuloc, line 35: we removed this because it does not render well on Github sadly ...so need to put the full command there .

docs/advanced_usages/caching.md
```diff
@@ -0,0 +1,85 @@
...
+      ranges:
+      - start: 172.22.22.10
+        end: 172.22.22.19
+        disable_cache: True
+    ```
+    This approach effectively disables caching for **ALL** commands sent to devices targeted by the `disable_cache` key.
+
+3. For tests developpers, caching can be disabled for a specific [`AntaCommand`](../advanced_usages/as-python-lib.md#antacommand-class) or [`AntaTemplate`](../advanced_usages/as-python-lib.md#antatemplate-class) by setting the `use_cache` attribute to `False`. That means the command output will always be collected on the device and therefore, never use caching.
+
+    Also, since caching is implemented at the `AntaDevice` abstract class level, all subclasses will inherit that default behavior. As a result, if you need to disable caching in any custom implementation of `AntaDevice` outside of the ANTA framework, you must initialize `AntaDevice` with `disable_cache` set to `True`:
```
gmuloc, line 76: Let's add a subtitle like `### Prevent the use of cache completely in a child class of AntaDevice`

tests/units/test_device.py
```diff
@@ -92,3 +95,141 @@ def test__init__(self, device_data: dict[str, Any]) -> None:
...
+        },
+        "command": {
+            "command": "show version",
+            "use_cache": False,
+        },
+        "cache_hit": "unused",
+    },
+]
+
+pytest_plugins = ("pytest_asyncio",)
```
gmuloc, line 179: We should move this somewhere else -will take care of this :)

titom73 (approved):

titom73 (approved):
Looks good after resolving points raised by @gmuloc

Issue #355, opened 2023-08-09
Some `anta get` commands do not need mandatory options in `anta`


mtache:
User should be able to use `anta get from-cvp` or `anta get from-ansible` without providing mandatory attributes in `anta`

PR #469, opened 2023-11-16
fix(anta.cli): Use configured anta inventory as output by default in get from-ansible

titom73:
# Description

Use configured anta inventory if `--output` is not configured in `anta get from-ansible` command.

```bash
anta get from-ansible --ansible-inventory ../anta-lab/atd-inventory/inventory.yml
[19:02:09] INFO     Building inventory from ansible file ../anta-lab/atd-inventory/inventory.yml
[19:02:12] INFO     output anta inventory is: anta-lab/network-tests/inventory.yml
           INFO        * adding entry for cv_atd1
           INFO        * adding entry for spine1
           INFO        * adding entry for spine2
           INFO        * adding entry for leaf1 
           INFO        * adding entry for leaf2 
           INFO        * adding entry for leaf3
           INFO        * adding entry for leaf4
           INFO     ANTA device inventory file has been created in anta-lab/network-tests/inventory.yml
```

If destination file has content, then cli asks confirmation to user if he wants to overwrite:

```bash
anta get from-ansible --ansible-inventory ../anta-lab/atd-inventory/inventory.yml
[19:07:19] INFO     Building inventory from ansible file ../anta-lab/atd-inventory/inventory.yml
Your destination file (anta-lab/network-tests/inventory.yml) is not empty, continue? [y/n]: n
[19:07:27] CRITICAL conversion aborted since destination file is not empty 
```

> If `anta` is executed in a non interactive environment, it breaks if destination file has content

To run script in a non-interactive environment, you can use `--overwrite` to tell anta it can overwrite file with content.

```bash
❯ anta get from-ansible --help
Usage: anta get from-ansible [OPTIONS]

  Build ANTA inventory from an ansible inventory YAML file

Options:
  -g, --ansible-group TEXT      Ansible group to filter
  -i, --ansible-inventory FILE  Path to your ansible inventory file to read
  -o, --output FILE             Path to save inventory file. If not
                                configured, use anta inventory file
  --overwrite                   Confirm script can overwrite existing
                                inventory file  [env var:
                                ANTA_GET_FROM_ANSIBLE_OVERWRITE]
  --help                        Show this message and exit.
```

# Checklist:

### DIFF
--- a/anta/cli/get/commands.py
+++ b/anta/cli/get/commands.py
@@ -118,12 +118,12 @@ def from_ansible(ctx: click.Context, output: Path, ansible_inventory: Path, ansi
     output = output if output is not None else ctx.obj["inventory_path"]
     anta_inventory_number_lines = 0
     if output.exists():
-        with open(output, "rbU") as f:
+        with open(output, "r", encoding="utf-8") as f:
             anta_inventory_number_lines = sum(1 for _ in f)
 
     if anta_inventory_number_lines > 0 and no_overwrite:
         logger.critical("conversion aborted since destination file is not empty and --no-overwrite is set")
-        sys.exit(ExitCode.INTERNAL_ERROR)
+        sys.exit(ExitCode.USAGE_ERROR)
 
     if anta_inventory_number_lines > 0 and not confirm_overwrite:
         confirm_overwrite = Confirm.ask(f"Your destination file ({output}) is not empty, continue?")
--- a/anta/cli/get/commands.py
+++ b/anta/cli/get/commands.py
@@ -86,8 +86,8 @@ def from_cvp(inventory_directory: str, cvp_ip: str, cvp_username: str, cvp_passw
 @click.option(
     "--output",
     "-o",
-    default="inventory-ansible.yml",
-    help="Path to save inventory file",
+    required=False,
+    help="Path to save inventory file. If not configured, use anta inventory file",
     type=click.Path(file_okay=True, dir_okay=False, exists=False, writable=True, path_type=Path),
 )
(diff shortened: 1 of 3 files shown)


titom73:
Need to discuss about how to manage a change of the default inventory file
provided

This change is breaking CI in the second run

anta/cli/get/commands.py
```diff
@@ -86,16 +87,54 @@ def from_cvp(inventory_directory: str, cvp_ip: str, cvp_username: str, cvp_passw
 @click.option(
     "--output",
     "-o",
-    default="inventory-ansible.yml",
-    help="Path to save inventory file",
+    required=False,
+    help="Path to save inventory file. If not configured, use anta inventory file",
```
gmuloc, line 91: this means that we still "REQUIRE" an anta-inventory configured even when output is given - which is very weird usage .. I am not sure I like this

@mtache  thoughts?

mtache: Agree. We need to merge #447 first.
As reported in #355, one current issue is that we require the user to define an inventory path for the command that generate this inventory: the workflow is broken.

Then, this PR can be implemented by using the inventory path as output if the user does not provide it as parameter, and fail the command if the user has not provided the output parameter nor the inventory path.

mtache (changes requested):

anta/cli/get/commands.py
titom73: Logic has been revamped to support PR #447 so, 

- if `--output` is defined we take it first
- if `anta --inventory` is defined and not `--output` we take it
- if `anta --inventory` is defined or not `--output` are not defined, cli fails

Also, only `--overwrite` remains. If not set and cli is running in an interactive TTY, we prompt user, else we break

PR #480, opened 2023-11-24
feat: Custom type Interface will format interface name

mtache:
# Description

Fixes # (issue id)

# Checklist:

### DIFF
--- a/anta/custom_types.py
+++ b/anta/custom_types.py
@@ -4,10 +4,11 @@
 """
 Module that provides predefined types for AntaTest.Input instances
 """
+import re
 from typing import Literal
 
 from pydantic import Field
-from pydantic.functional_validators import AfterValidator
+from pydantic.functional_validators import AfterValidator, BeforeValidator
 from typing_extensions import Annotated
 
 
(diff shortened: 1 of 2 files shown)


anta/custom_types.py
```diff
@@ -17,13 +18,40 @@ def aaa_group_prefix(v: str) -> str:
...
+def interface_autocomplete(v: str) -> str:
+    """Allow the user to only provide the beginning of an interface name
+    E.g. et, eth will be changed to Ethernet"""
+    intf_id_re = re.compile(r"[0-9]+(\/[0-9]+)*(\.[0-9]+)?")
+    m = intf_id_re.search(v)
+    if m is None:
+        raise ValueError(f"Could not parse interface ID in interface '{v}'")
+    intf_id = m[0]
+    if any(v.lower().startswith(p) for p in ["et", "eth"]):
+        return f"Ethernet{intf_id}"
```
gmuloc, line 30: Should we document this?
Also it only works for ethernet interfaces in the end for now right? Would we not rather force people to have nice catalog inputs rather than taking the load?
(I agree with keep case sensitivity - but for instance what do we do with Port-Channels / Port-channel / Port-ChaNNel?)

anta/custom_types.py
```diff
@@ -17,13 +18,40 @@ def aaa_group_prefix(v: str) -> str:
...
 TestStatus = Literal["unset", "success", "failure", "error", "skipped"]
 
 # AntaTest.Input types
 AAAAuthMethod = Annotated[str, AfterValidator(aaa_group_prefix)]
 Vlan = Annotated[int, Field(ge=0, le=4094)]
 Vni = Annotated[int, Field(ge=1, le=16777215)]
-Interface = Annotated[str, Field(pattern=r"^(Ethernet|Fabric|Loopback|Management|Port-Channel|Tunnel|Vlan|Vxlan)[0-9]+(\/[0-9]+)*(\.[0-9]+)?$")]
+Interface = Annotated[
+    str,
+    Field(pattern=r"^(Ethernet|Fabric|Loopback|Management|Port-Channel|Tunnel|Vlan|Vxlan)[0-9]+(\/[0-9]+)*(\.[0-9]+)?$"),
```
gmuloc, line 52: Let's get prepared for WAN ;) 
```suggestion
    Field(pattern=r"^(Dps|Ethernet|Fabric|Loopback|Management|Port-Channel|Tunnel|Vlan|Vxlan)[0-9]+(\/[0-9]+)*(\.[0-9]+)?$"),
```

gmuloc (changes requested):

anta/custom_types.py
mtache: actually, we support "Port-ChaNNel" now :)
For the rest, I could see this coming so I've implemented it...
But we could support "po", similar as what is done in EOS.

mtache: added doc

gmuloc (approved):

PR #532, opened 2024-01-18
feat(anta.tests): Added testcase for MLAG primary priority

### LINKED ISSUE #529: Add the testcase to verify MLAG priority feature

MaheshGSLAB:
# Description

Added testcase for MLAG primary priority

Fixes #529 
Success: The test will pass if the MLAG state is set as 'primary' and the priority matches the input.
Failure: The test will fail if the MLAG state is not 'primary' or the priority doesn't match the input.
Skipped: The test will be skipped if MLAG is 'disabled'.

# Checklist:

### DIFF
--- a/anta/tests/mlag.py
+++ b/anta/tests/mlag.py
@@ -10,8 +10,10 @@ from __future__ import annotations
 
 from pydantic import conint
 
+from anta.custom_types import MlagPriority
 from anta.models import AntaCommand, AntaTest
 from anta.tools.get_value import get_value
+from anta.tools.utils import get_failed_logs
 
 
 class VerifyMlagStatus(AntaTest):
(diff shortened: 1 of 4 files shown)


carl-baillargeon (approved):
LGTM!

anta/tests/mlag.py
```diff
@@ -192,3 +194,49 @@ def test(self) -> None:
...
+
+        # Define the expected MLAG details
+        expected_mlag_details = {"mlagState": "primary", "primaryPriority": self.inputs.primary_priority}
+
+        # Check if the actual MLAG details match the expected details
+        if actual_mlag_details == expected_mlag_details:
+            self.result.is_success()
+        else:
+            failed_log = get_failed_logs(expected_mlag_details, actual_mlag_details)
+            self.result.is_failure(f"The device is not set as the MLAG primary or the primary priority does not match the expected value:{failed_log}")
```
gmuloc, line 242: I would rather have a clear message that says in one case `the devcie is not set as primary` and another case `the primary priority does not match expcted`

This is quite straighforward and gives better and succint information + if these are two messages then we could technicaly separate them later in the lists of error messages

titom73 (approved):

PR #537, opened 2024-01-22
feat(anta.tests): Added testcase to verify DNS ip name server

### LINKED ISSUE #535: Add the testcase to verify DNS status

MaheshGSLAB:
# Description

Added testcase to verify DNS ip name server

Fixes #535 
* success: The test will pass if the DNS server specified in the input is configured with the correct VRF and priority.
* failure: The test will fail if the DNS server is not configured or if the VRF and priority of the DNS server do not match the input.

# Checklist:


anta/tests/services.py
```diff
@@ -0,0 +1,72 @@
...
+        self.result.is_success()
+        for server in self.inputs.dns_servers:
+            address = str(server.server_address)
+            vrf = server.vrf
+            priority = server.priority
+            input_dict = {"ipAddr": address, "vrf": vrf}
+
+            if get_item(command_output, "ipAddr", address) is None:
+                self.result.is_failure(f"DNS server `{address}` is not configured with any VRF.")
+                continue
```
carl-baillargeon, line 65: Do we need that? Would it be possible to create a single failure message with the input DNS servers not configured properly?

carl-baillargeon (changes requested):
Thanks Mahesh. Since services.py is a new module, you need to create the documentation: anta/docs/api/tests.services.md and add the section in anta/docs/api/tests.md.

Finally add your newly created MD file to mkdocs.yml. Double check the doc with mkdocs serve.

Thanks

anta/tests/services.py
MaheshGSLAB: it is possible but then we will not able to get the detail that what is not configured properly. There are two case, first one is that DNS sever is not configured with any of the VRF and the other case is DNS server is configured but not with the expected vrf.

carl-baillargeon (approved):
LGTM

PR #601, opened 2024-03-24
fix(anta): Fix anta exec clear-counters by adding supported platforms

### LINKED ISSUE #513: `clear hardware counter drop` is not supported on some platforms
Update `anta exec clear-counters` script to avoid using `clear hardware counter drop` on unsupported platforms.

carl-baillargeon:
# Description

Add platform_utils.py to be able to properly run or skip tests.

Fixes #513 #607 

# Checklist:

- [ ] Fix pytest
- [ ] Update doc

### DIFF
--- a/anta/device.py
+++ b/anta/device.py
@@ -343,7 +343,7 @@ class AsyncEOSDevice(AntaDevice):
             )
         except aioeapi.EapiCommandError as e:
             command.errors = e.errors
-            if self.supports(command):
+            if self.supports(command, log=True):
                 logger.error("Command '%s' failed on %s", command.command, self.name)
         except (HTTPError, ConnectError) as e:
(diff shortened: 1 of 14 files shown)


anta/device.py
```diff
@@ -171,12 +171,13 @@ async def collect_commands(self, commands: list[AntaCommand]) -> None:
         """
         await asyncio.gather(*(self.collect(command=command) for command in commands))
 
-    def supports(self, command: AntaCommand) -> bool:
+    def supports(self, command: AntaCommand, *, log: bool = True) -> bool:
```
mtache, line 174: Any reason why using a boolean for logging and set it to True only once ?
I would suggest to remove this boolean and put the warning where it was :)

PR #631, opened 2024-04-16
ci: Remove upper bounds for requirements

gmuloc:
# Description

Drinking the cool aid from here: https://iscinumpy.dev/post/bound-version-constraints/
ANTA is intended to be used as a library

* Removed upper cap on requirements (not dev)
* Add warnings as error in pytest as recommended
    * ignoring the one from cvprac (need to see if we can fix)
    * and 3 pytest warnings that require investigation

# TODO

- [x] Decide if we want to add a lockfile or to recommend people to use pipx when installing ANTA to be used as an App.
- [x] Do not merge - doc needed

# Checklist:


carl-baillargeon (dismissed):
It works fine in a new environment but the documentation is not updated :)

mtache (commented):
Should we advise to use pipx when using ANTA with the CLI ?

mtache (approved):

PR #634, opened 2024-04-18
feat: Add feature to search config

### LINKED ISSUE #632: Feature: Add support for generic checks in running config.
As an engineer writing custom test catalogs, I would find it useful to be able to check that certain lines are in the running config, especially if there are no existing tests for the functionality I am interested in.

Example: 
- Checking that the enable password is set
- Checking for a particular username has been set  in the config.

Probably would need to support a regex search or searching for line that include the given text.

chetryan:
# Description

Add an Anta Test to search the running configuration for a list of regular expressions.

Fixes #632 

# Checklist:

### DIFF
--- a/anta/tests/configuration.py
+++ b/anta/tests/configuration.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import re
 from typing import TYPE_CHECKING, ClassVar
 
+from anta.custom_types import RegexString
 from anta.models import AntaCommand, AntaTest
 
 if TYPE_CHECKING:
@@ -81,9 +82,11 @@ class VerifyRunningConfigDiffs(AntaTest):
 class VerifyRunningConfigLines(AntaTest):
     """Verifies the given regular expression patterns are present in the running-config.
 
-    **Since this uses regular expression searches on the whole running-config, it can
-    drastically impact performance and should only be used if no other tests is available for.**
-    If possible, try using another Anta test that is more specific.
+    !!! warning
+        Since this uses regular expression searches on the whole running-config, it can
+        drastically impact performance and should only be used if no other test is available.
+
+        If possible, try using another ANTA test that is more specific.
 
     Expected Results
     ----------------
@@ -109,7 +112,7 @@ class VerifyRunningConfigLines(AntaTest):
     class Input(AntaTest.Input):
         """Input model for the VerifyRunningConfigLines test."""
 
-        regex_patterns: list[str]
+        regex_patterns: list[RegexString]
         """List of regular expressions."""
 
     @AntaTest.anta_test
--- a/anta/custom_types.py
+++ b/anta/custom_types.py
@@ -81,6 +81,16 @@ def bgp_multiprotocol_capabilities_abbreviations(value: str) -> str:
     return value
 
 
+def validate_regex(value: str) -> str:
+    """Validate that the input value is a valid regex format."""
+    try:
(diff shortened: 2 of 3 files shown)


anta/tests/configuration.py
```diff
@@ -75,3 +76,55 @@ def test(self) -> None:
...
         else:
             self.result.is_failure(command_output)
+
+
+class VerifyRunningConfigLines(AntaTest):
+    """Verifies the given regular expression patterns are present in the running-config.
+
+    **Since this uses regular expression searches on the whole running-config, it can
+    drastically impact performance and should only be used if no other tests is available for.**
+    If possible, try using another Anta test that is more specific.
```
carl-baillargeon, line 86:
```suggestion
    !!! warning
    Since this uses regular expression searches on the whole running-config, it can
    drastically impact performance and should only be used if no other tests is available for.
    If possible, try using another ANTA test that is more specific.
```

carl-baillargeon: If you do `mkdocs serve` locally you can see the documentation. `!!! warning` in `mkdocs` adds a nice little box.

anta/tests/configuration.py
```diff
@@ -75,3 +76,55 @@ def test(self) -> None:
...
+        """List of regular expressions."""
+
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """Main test function for VerifyRunningConfigLines."""
+        failure_msgs = []
+        command_output = self.instance_commands[0].text_output
+
+        for pattern in self.inputs.regex_patterns:
+            re_search = re.compile(pattern, flags=re.MULTILINE)
```
carl-baillargeon, line 122: This line here will generate an error if the regex string is invalid (e.g. `'['`). We can catch the error when ANTA parses the inputs with a custom type.

In the `anta.custom_types` module, add something like this:
```
def validate_regex(value: str) -> str:
    """Validate that the input value is a valid regex format."""
    try:
        re.compile(value)
    except re.error as e:
        msg = f"Invalid regex: {e}"
        raise ValueError(msg) from e
    return value
 ```
 And then create the custom type in the same file: `RegexString = Annotated[str, AfterValidator(validate_regex)]`
 
 You can now use that custom type on line 112 above.
 
With that in place we can catch a bad regex from the start.

File: anta/tests/configuration.py [hunk 3]
```diff
@@ -75,3 +76,55 @@ def test(self) -> None:
...
+
+    name = "VerifyRunningConfigLines"
+    description = "Search the Running-Config for the given RegEx patterns."
+    categories: ClassVar[list[str]] = ["configuration"]
+    commands: ClassVar[list[AntaCommand | AntaTemplate]] = [AntaCommand(command="show running-config", ofmt="text")]
+
+    class Input(AntaTest.Input):
+        """Input model for the VerifyRunningConfigLines test."""
+
+        regex_patterns: list[str]
```
carl-baillargeon, line 112:
```suggestion
        regex_patterns: list[RegexString]
```
See my comment below

anta/tests/configuration.py
chetryan: This is great. I did try to find a way to only compile the regex once, however that wouldn't be a validator. At the end I went your suggestion.

gmuloc (approved):
LGTM - thanks @chetryan

carl-baillargeon (approved):
LGTM. Thanks

PR #654, opened 2024-04-24
refactor(anta): Optimize AntaTemplate render performance.

### LINKED ISSUE #647: ANTA runner performance issues at high scale
Tested ANTA on a fabric of **1040** devices and **2 606 033** tests.  ANTA was run outside of the CLI, meaning no CLI tags were used. The catalog consisted of evenly distributed tests across devices, around 75% of the tests had filters (tags) for specific devices.

Used  [tqdm](https://tqdm.github.io/) to monitor iterations per second (it/s) on various loops in the runner and cProfile.

#### Findings:
1. Creating AntaTestRunner tuples from the tags `for device in inventory.devices:` --> 12 it/s (3h24m total)
    a. `selected_tests.update((t, device) for t in catalog.tests if t.inputs.filters is None or t.inputs.filters.tags is None)` --> 860 000 it/s (3-4 seconds per device) 
    b. `selected_tests.update((t, device) for t in catalog.get_tests_by_tags(device.tags))` --> 320 000 it/s (7-8 seconds per device)
    * addressed by #648
    
2. Creating coroutines `for test_definition, device in selected_tests:` --> 294 it/s (2h27m total)
    a. Looking at cProfile, most of time is spent in Pydantic, which makes sense since `AntaCommand` and `AntaTemplate` are Pydantic models.
    * to be addressed
    
    Here is a screenshot of the runner cProfile based on 45 000 tests:
image

#### Other findings:
1. `await selected_inventory.connect_inventory()` Here we connect to all devices of the inventory even though we might not run tests on these devices depending on the tags from the catalog/inventory.
    * to be addressed

gmuloc:
# Description

Builds on top of:
* #648
* #650

Adds:

* cprofile decorator around `anta.runner.main`
* Do not create a pydantic model at every AntaTemplate.render() call. Instead, create a `ParamsSchema` model in AntaTemplate constructor where we can infer the fields.

Fixes #647

# Checklist:


anta/models.py
```diff
@@ -71,11 +101,37 @@ class AntaTemplate(BaseModel):
...
     """
 
-    template: str
-    version: Literal[1, "latest"] = "latest"
-    revision: Revision | None = None
-    ofmt: Literal["json", "text"] = "json"
-    use_cache: bool = True
+    # pylint: disable=too-few-public-methods
+
+    __metaclass__ = SingletonArgs
```
gmuloc, line 106: should probably be `class AntaTemplate(metaclass=SingletonArgs)`

gmuloc: so yes it SHOULD (as `__metadata__` does not work anymore)
but also it is useless so it was
1. fixed in [350546d](https://github.com/arista-netdevops-community/anta/pull/654/commits/350546dfe6875c7aea77682e8fb4100ca514b719) _if we ever need this later_
2. removed in [96dc467](https://github.com/arista-netdevops-community/anta/pull/654/commits/96dc467a5341956cf4f881f770db1a9d685112f9)

anta/models.py
```diff
@@ -88,34 +104,28 @@ def render(self, **params: str | int | bool) -> AntaCommand:
...
+            command = self.template.format(**params)
+        except (KeyError, SyntaxError) as e:
             raise AntaTemplateRenderError(self, e.args[0]) from e
+        return AntaCommand(
+            command=command,
+            ofmt=self.ofmt,
+            version=self.version,
+            revision=self.revision,
+            template=self,
+            params=self.params_schema(**params),
```
gmuloc, line 126: this could raise with extra Key - don't we want the AntaTemplateRenderError in that case?

gmuloc: hmm I see the test is looking at the pydantic error. This is actually an error for test developers so probably good as it is. Let's leave it as it.

PR #657, opened 2024-04-26
feat(anta): add AntaCatalog.dump() and AntaCatalog.merge()

mtache:
- **feat: add AntaCatalog.dump() and AntaCatalog.merge()**
- **ci: add unit tests for anta.catalog module**

# Description

This PR also includes functions to merge AntaCatalog instances and dump as a YAML file:
- `AntaCatalog.merge() -> AntaCatalog`
- `AntaCatalog.dump() -> AntaCatalogFile`
- `AntaCatalogFile.yaml()`

Script example:
``` python
#!/usr/bin/env python
from anta.catalog import AntaCatalog

from pathlib import Path
from anta.models import AntaTest


CATALOG_SUFFIX = '-catalog.yml'
CATALOG_DIR = 'intended/test_catalogs/'

if __name__ == "__main__":
    catalog = AntaCatalog()
    for file in Path(CATALOG_DIR).glob('*'+CATALOG_SUFFIX):
        c = AntaCatalog.parse(file)
        device = str(file).removesuffix(CATALOG_SUFFIX).removeprefix(CATALOG_DIR)
        print(f"Merging test catalog for device {device}")
        # Apply filters to all tests for this device
        for test in c.tests:
            test.inputs.filters = AntaTest.Input.Filters(tags=[device])
        catalog.merge(c)
    with open(Path('anta-catalog.yml'), "w") as f:
        f.write(catalog.dump().yaml())

```

# Checklist:


anta/catalog.py
```diff
@@ -94,8 +111,7 @@ def instantiate_inputs(
                 f"{test_class.name} test inputs are not valid: {inputs_msg}\n",
                 {"errors": e.errors()},
             ) from e
-        msg = f"Could not instantiate inputs as type {type(data).__name__} is not valid"
-        raise ValueError(msg)
+        return data
```
gmuloc, line 114: Hmmm you are changing the output if data is not of correct type - no more meaningful error message?

anta/catalog.py
```diff
@@ -182,6 +198,9 @@ def check_tests(cls: type[AntaCatalogFile], data: Any) -> Any:  # noqa: ANN401
             for module, tests in typed_data.items():
                 test_definitions: list[AntaTestDefinition] = []
                 for test_definition in tests:
+                    if isinstance(test_definition, AntaTestDefinition):
+                        test_definitions.append(test_definition)
+                        continue
```
gmuloc, line 203: Good catch ..

anta/catalog.py
mtache: The` raise ValueError` was dead code. If I remove it, pylint complains about returns statements not being consistent, so I'm returning the data here assigned in the try block.
If you end up in the except block, you won't reach this return statement.

anta/catalog.py
```diff
@@ -335,6 +367,34 @@ def from_list(data: ListAntaTestTuples) -> AntaCatalog:
...
+        Args:
+        ----
+            catalog: AntaCatalog instance to merge to this instance.
+
+        Returns
+        -------
+            A new AntaCatalog instance containing the tests of the two instances.
+        """
+        tests = self.tests
+        tests.extend(catalog.tests)
```
gmuloc, line 382: are you not modifying in place `self.tests` here?

```
>> class Test:
...   def __init__(self, tests: list|None = None):
...     self.tests = tests or []
...   def merge(self, other):
...     tests = self.tests
...     tests.extend(other.tests)
...     return Test(tests=tests)
...
>>> a = Test()
>>> b = Test([1,2,3])
>>> a.tests
[]
>>> b.tests
[1, 2, 3]
>>> c = a.merge(b)
>>> c.tests
[1, 2, 3]
>>> a.tests
[1, 2, 3]
```

anta/catalog.py
```diff
@@ -335,6 +367,34 @@ def from_list(data: ListAntaTestTuples) -> AntaCatalog:
...
+    def dump(self) -> AntaCatalogFile:
+        """Return an AntaCatalogFile instance from this AntaCatalog instance.
+
+        Returns
+        -------
+            An AntaCatalogFile instance containing tests of this AntaCatalog instance.
+        """
+        root: dict[ImportString[Any], list[AntaTestDefinition]] = {}
+        for test in self.tests:
+            module = test.test.__module__
```
gmuloc, line 394: Should we add a `module` property in AntaTest to pick this up nicely?

tests/data/test_inventory_medium.yml
```diff
@@ -0,0 +1,14 @@
+anta_inventory:
+  hosts:
+  - host: clab-evpn-vxlan-fabric-spine1
+    name: mtache-spine1
```
gmuloc, line 4: want to keep mtache?

tests/units/test_catalog.py
```diff
@@ -312,9 +312,47 @@ def test_build_indexes_filtered(self) -> None:
...
         catalog.build_indexes()
         tests: set[AntaTestDefinition] = catalog.get_tests_by_tags(tags={"leaf"})
         assert len(tests) == 3
         tests = catalog.get_tests_by_tags(tags={"leaf", "spine"}, strict=True)
         assert len(tests) == 1
+
+    def test_merge(self) -> None:
+        """Test AntaCatalog.merge()."""
+        catalog1: AntaCatalog = AntaCatalog.parse(DATA_DIR / "test_catalog.yml")
+        assert len(catalog1.tests) == 1
```
gmuloc, line 325: add assertion to check original catalog did not change :p

gmuloc (changes requested):

tests/data/test_inventory_medium.yml
mtache: Good catch :)

anta/catalog.py
mtache: I have defined the property in `AntaTest` ABC but this cannot be used before the class is actually instanciated (which is not the case here). I don't have better ideas.

anta/catalog.py
gmuloc: it is not deadcode if you give data=wrong type as input.

gmuloc: Adding a print statement above the return `data`
```
>>> AntaTestDefinition(test = anta.models.AntaTest, inputs=[])
WOUHOU I AM OUT OF TRY EXCEPT RETURNING [] of type <class 'list'>
  File "<stdin>", line 1, in <module>
  File "/Users/guillaumemulocher/Github/gmuloc/anta/anta/catalog.py", line 70, in __init__
    self.__pydantic_validator__.validate_python(
pydantic_core._pydantic_core.ValidationError: 1 validation error for AntaTestDefinition
inputs
  Input should be a valid dictionary or instance of Input [type=model_type, input_value=[], input_type=list]
    For further information visit https://errors.pydantic.dev/2.6/v/model_type
 ```

**Review by gmuloc — COMMENTED** [review]
I think we should put back the ValueError we had as it is a bit clearer than the cryptic pydantic error.

File: anta/catalog.py (hunk 1 above)
Comment (mtache) [reply anta/catalog.py:115]: Ok this was not dead code.
But you got a ValidationError, so what's your point ?

Comment (mtache) [reply anta/catalog.py:115]: It just changes the error message.
The previous one was more or less the same level of info. The default from pydantic indicates the correct type to use:
```
>>> AntaTestDefinition(test = anta.models.AntaTest, inputs=[])
  File "<stdin>", line 1, in <module>
  File "/Users/mtache/Git/projects/anta/anta/catalog.py", line 70, in __init__
    self.__pydantic_validator__.validate_python(
pydantic_core._pydantic_core.ValidationError: 1 validation error for AntaTestDefinition
inputs
  Value error, Could not instantiate inputs as type list is not valid [type=value_error, input_value=[], input_type=list]
    For further information visit https://errors.pydantic.dev/2.7/v/value_error
```

gmuloc (approved):

PR #660, opened 2024-04-29
test: More CLI coverage

gmuloc:
# Description

Fixes # (issue id)

# Checklist:

### DIFF
--- a/anta/cli/debug/utils.py
+++ b/anta/cli/debug/utils.py
@@ -12,7 +12,6 @@ from typing import TYPE_CHECKING, Any, Callable
 import click
 
 from anta.cli.utils import ExitCode, inventory_options
-from anta.logger import anta_log_exception
 
 if TYPE_CHECKING:
     from anta.inventory import AntaInventory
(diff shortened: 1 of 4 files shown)


anta/cli/debug/utils.py
```diff
@@ -54,8 +55,8 @@ def wrapper(
         try:
             d = inventory[device]
         except KeyError as e:
-            message = f"Device {device} does not exist in Inventory"
-            logger.error(e, message)
+            message = f"Device '{device}' does not exist in Inventory"
+            anta_log_exception(e, message)
```
mtache, line 59: Not really what we want.
`anta_log_exception` print a traceback to debug.
This except block does not need to be debug.
It is more something like
```
if device not in inventory:
   logger.error(f"Device {device} does not exist in Inventory")
   ctx.exit(ExitCode.USAGE_ERROR)
```

That we need.

mtache (dismissed):

anta/cli/debug/utils.py
gmuloc: `anta_log_exception` does not `print a traceback to debug`. It adds a log for the exception either in `critical` or if ANTA_DEBUG is set, it prints the traceback (which granted we don't need in this case)
https://github.com/arista-netdevops-community/anta/blob/6b7a54ff93c76250d242d3796f6cc84bb51813e0/anta/logger.py#L116 (critical if no ANTA_DEBUG set, exception otherwise)

gmuloc: ok that was not the point - we should not raise ;) coffee needs more ☕

PR #672, opened 2024-05-13
feat(anta): Add support of CSV file export

titom73:
# Description

Add option to save tests result in CSV format

## Command example

```bash
# Example 1
anta nrfu -c anta/nrfu.yml csv --csv-output demo.csv

# Example 2
anta nrfu -c anta/nrfu.yml --device spine1 csv --csv-output demo.csv
```

image

# Checklist:

# Requirements

- [x] PR #750 before updating documentation with `anta nrfu --help output`


anta/cli/nrfu/commands.py
```diff
@@ -63,6 +60,28 @@ def text(ctx: click.Context) -> None:
...
+    type=click.Path(
+        file_okay=True,
+        dir_okay=False,
+        exists=False,
+        writable=True,
+        path_type=pathlib.Path,
+    ),
+    show_envvar=True,
+    required=False,
+    help="Path to save report as a CSV file. It only saves test results and not the output from --group-by option",
```
carl-baillargeon, line 76: Do we still need "It only saves test results and not the output from --group-by option" with your latest changes?

anta/cli/nrfu/commands.py
```diff
@@ -63,6 +60,28 @@ def text(ctx: click.Context) -> None:
     exit_with_code(ctx)
 
 
+@click.command()
+@click.pass_context
+@click.option(
+    "--csv-output",
```
carl-baillargeon, line 66: Should we make this more generic like `-o or --output` like the JSON option? This is also what I did in my MD PR but open to change it.

anta/cli/nrfu/utils.py
```diff
@@ -122,6 +124,14 @@ def print_jinja(results: ResultManager, template: pathlib.Path, output: pathlib.
             file.write(report)
 
 
+def save_to_csv(ctx: click.Context, csv_file: pathlib.Path | None = None) -> None:
+    """Save results to a CSV file."""
+    if csv_file is not None:
+        ReportCsv().csv_report(results=_get_result_manager(ctx), csv_filename=csv_file)
```
carl-baillargeon, line 130: More appropriate to use a classmethod to generate the report since we don't really need to keep state with an instance;

`ReportCsv.generate()` or `ReportCsv.csv_report()`, I used `MDReport.generate_report()` for the MD report. Again willing to change it but we should be consistent.

I think `CSVReport.generate()` and `MDReport.generate()` looks appropriate.

anta/reporter/__init__.py
```diff
@@ -27,6 +28,18 @@
 class ReportTable:
     """TableReport Generate a Table based on TestResult."""
 
+    class Headers(BaseModel):
```
carl-baillargeon, line 31: Should we use a `dataclass` here to be able to use like `headers.number_of_success`? I think it would be cleaner.

anta/reporter/__init__.py
```diff
@@ -254,3 +264,73 @@ def render(self, data: list[dict[str, Any]], *, trim_blocks: bool = True, lstrip
             template = Template(file_.read(), trim_blocks=trim_blocks, lstrip_blocks=lstrip_blocks)
 
         return template.render({"data": data})
+
+
+class ReportCsv:
```
carl-baillargeon, line 269: Duplicate code? Since you moved it to a seperate module.

anta/reporter/csv_reporter.py
```diff
@@ -0,0 +1,90 @@
...
+    from anta.result_manager import ResultManager
+    from anta.result_manager.models import TestResult
+
+logger = logging.getLogger(__name__)
+
+
+class ReportCsv:
+    """Build a CSV report."""
+
+    class Headers(BaseModel):
```
carl-baillargeon, line 26: Yeah I think a dataclass would be cleaner. We can put it in a models.py in either reporter or result_manager package and we can reuse it.

carl-baillargeon: (I would reuse in the MD report since they are the same)

anta/cli/nrfu/commands.py
titom73: I am just thinking it might be untrivial once we move to multiple commands approach having the same knob for different commands

anta/cli/nrfu/commands.py
titom73: It is still the case: we don't save `--group-by` results. So the warning message

anta/reporter/__init__.py
titom73: Yes true. Code removed

anta/reporter/__init__.py
titom73: Convert all `class Headers` to `dataclass`. But still per report type since there is no common headers across all reports

anta/reporter/csv_reporter.py
titom73: Convert all class Headers to dataclass. But still per report type since there is no common headers across all reports

anta/cli/nrfu/utils.py
titom73: Update code accordingly: 

```python
ReportCsv.generate(results=_get_result_manager(ctx), csv_filename=csv_file)`
```

docs/cli/nrfu.md
```diff
@@ -145,6 +145,29 @@ anta nrfu --tags LEAF json
 ```
 ![$1anta nrfu json results](../imgs/anta-nrfu-json-output.png){ loading=lazy width="1600" }
 
+## Performing NRFU with saving results in CSV file.
```
carl-baillargeon, line 148:
```suggestion
## Performing NRFU and saving results in a CSV file.
```

docs/cli/nrfu.md
```diff
@@ -145,6 +145,29 @@ anta nrfu --tags LEAF json
...
+
+The `csv` command in NRFU testing is useful in generating a CSV file with all tests result that can be analyzed, filtered by operator for reporting purposes.
```
carl-baillargeon, line 150:
```suggestion
The `csv` command in NRFU testing is useful for generating a CSV file with all tests result. This file can be easily analyzed and filtered by operator for reporting purposes.
```

docs/snippets/anta_nrfu_help.txt
```diff
@@ -52,6 +52,7 @@ Options:
   --help                          Show this message and exit.
 
 Commands:
+  csv         ANTA command to check network states with CSV result.
```
carl-baillargeon, line 55:
```suggestion
  csv         ANTA command to check network state with CSV report.
```

tests/units/reporter/test_csv.py
```diff
@@ -0,0 +1,82 @@
...
+
+# pylint: disable=too-few-public-methods
+
+import csv
+import pathlib
+
+from anta.reporter.csv_reporter import ReportCsv
+from anta.result_manager import ResultManager
+
+# To avoid succh error:
```
carl-baillargeon, line 14:
```suggestion
# To avoid such error:
```

anta/reporter/csv_reporter.py
```diff
@@ -0,0 +1,92 @@
...
+        test_status: str = "Test Status"
+        messages: str = "Message(s)"
+        description: str = "Test description"
+        categories: str = "Test category"
+
+    @classmethod
+    def _split_list_to_txt_list(cls, usr_list: list[str], delimiter: str | None = None) -> str:
+        """Split list to multi-lines string.
+
+        Args:
```
gmuloc, line 40: we use numpy style
```suggestion
       Parameters
```

docs/cli/nrfu.md
```diff
@@ -145,6 +145,29 @@ anta nrfu --tags LEAF json
...
+```bash
+anta nrfu csv --help
+Usage: anta nrfu csv [OPTIONS]
+
+  ANTA command to check network states with CSV result.
+
+Options:
+  --csv-output FILE  Path to save report as a CSV file. It only saves test
+                     results and not the output from --group-by option  [env
+                     var: ANTA_NRFU_CSV_CSV_OUTPUT]
```
gmuloc, line 163: I would move `It only saves test
                     results and not the output from --group-by option` to the command help rather than here

tests/units/reporter/test_csv.py
```diff
@@ -0,0 +1,82 @@
...
+            ),
+            FakeResult(
+                name="dummy",
+                test="VerifyHardwareStatus",
+                result="failure",
+                messages=["Test failed"],
+                description="Verify hardware status",
+                categories=["category1"],
+            ),
+        ]
```
gmuloc, line 46: we have pytest fixtures for this: https://github.com/aristanetworks/anta/blob/main/tests/lib/fixture.py#L144 you can use it in the test to avoid recreating everything.

gmuloc (changes requested):

tests/units/reporter/test_csv.py
titom73: Fixed

docs/cli/nrfu.md
titom73: Agreed, but it will be only visible with `anta nrfu csv --help`

```bash
❯ anta nrfu --help G csv
  csv         ANTA command to check network states with CSV result.
❯ anta nrfu csv --help
Usage: anta nrfu csv [OPTIONS]

  ANTA command to check network states with CSV result. It only saves test
  results and not the output from --group-by option.

Options:
  --csv-output FILE  Path to save report as a CSV file  [env var:
                     ANTA_NRFU_CSV_CSV_OUTPUT]
  --help             Show this message and exit.
```

anta/cli/nrfu/commands.py
```diff
@@ -63,6 +60,28 @@ def text(ctx: click.Context) -> None:
...
+        exists=False,
+        writable=True,
+        path_type=pathlib.Path,
+    ),
+    show_envvar=True,
+    required=False,
+    help="Path to save report as a CSV file",
+)
+def csv(ctx: click.Context, csv_output: pathlib.Path) -> None:
+    """ANTA command to check network states with CSV result. It only saves test results and not the output from --group-by option."""
```
gmuloc, line 79: so right now we don't support group-by on the csv command but only under the `table` command - does it make sense to keep this there? (I added the same comment in the MD report PR).

I guess it is because the target is to support multi CLI later (nrfu table and csv) ?

If so, it would look like the following I suppose: `anta nrfu table --group-by device csv --csv-output=toto.csv`  and so we would need to keep the comment?

anta/reporter/csv_reporter.py
```diff
@@ -0,0 +1,97 @@
...
+        headers = [
+            cls.Headers.device,
+            cls.Headers.test_name,
+            cls.Headers.test_status,
+            cls.Headers.messages,
+            cls.Headers.description,
+            cls.Headers.categories,
+        ]
+
+        with pathlib.Path.open(csv_filename, "w", encoding="utf-8") as csvfile:
```
gmuloc, line 90: this could probably fail and raise some OS error in some places right?

docs/cli/nrfu.md
gmuloc: yes but group-by is anyway only available using `anta nrfu table --help` so now I am even questioning adding this

tests/units/reporter/test_csv.py
```diff
@@ -0,0 +1,72 @@
...
+
+import csv
+import pathlib
+from typing import Any, Callable
+
+from anta.reporter.csv_reporter import ReportCsv
+from anta.result_manager import ResultManager
+
+# To avoid such error:
+# pytest.PytestCollectionWarning: cannot collect test class 'TestResult' because it has a __init__ constructor
```
gmuloc, line 16: I am not sure I understand the meaning of this comment - what is the problem here? I suppose it is because we have a class called TestResult in out code and pytest is doing some collection on it?

anta/reporter/csv_reporter.py
titom73: [`Pathlib`](https://docs.python.org/3/library/pathlib.html) is taking care of host filesystem, so I don't think it will raise some errors.

tests/units/reporter/test_csv.py
titom73: Comment that can be removed since there is no more `TestResult` import

anta/reporter/csv_reporter.py
gmuloc:
```
>>> p = pathlib.Path("/tmp/test_ro/ro_folder/toto.yml")
>>> with pathlib.Path.open(p, "w", encoding="utf-8")  as f:
...   f.write("blah")
...
  File "<stdin>", line 1, in <module>
  File "/Users/guillaumemulocher/.pyenv/versions/3.10.4/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/pathlib.py", line 1117, in open
    return self._accessor.open(self, mode, buffering, encoding, errors,
PermissionError: [Errno 13] Permission denied: '/tmp/test_ro/ro_folder/toto.yml'
```

not if it is a read only folder
we should catch OSError and raise a message indicating what goes wrong

titom73: got it

anta/cli/nrfu/utils.py
```diff
@@ -126,9 +127,13 @@ def print_jinja(results: ResultManager, template: pathlib.Path, output: pathlib.
 
 def save_to_csv(ctx: click.Context, csv_file: pathlib.Path) -> None:
     """Save results to a CSV file."""
-    ReportCsv.generate(results=_get_result_manager(ctx), csv_filename=csv_file)
-    checkmark = Emoji("white_check_mark")
-    console.print(f"CSV report saved to {csv_file} {checkmark}", style="cyan")
+    try:
+        ReportCsv.generate(results=_get_result_manager(ctx), csv_filename=csv_file)
+        checkmark = Emoji("white_check_mark")
+        console.print(f"CSV report saved to {csv_file} {checkmark}", style="cyan")
```
gmuloc, line 133:
```suggestion
        console.print(f"CSV report saved to {csv_file} ✅", style="cyan")
```

anta/cli/nrfu/utils.py
```diff
@@ -11,12 +11,15 @@
 from typing import TYPE_CHECKING, Literal
 
 import rich
+from rich.emoji import Emoji
```
gmuloc, line 14:
```suggestion
```

gmuloc (approved):

carl-baillargeon (approved):
LGTM

PR #680, opened 2024-05-16
feat(anta): New AntaRunner class with limit concurrency

carl-baillargeon:
# Description

- Add AntaRunner class
- Add settings.py module to manage ANTA settings
- Add Semaphore option to limit the number of tests running concurrently

Fixes: #832
Fixes: #1187 

**Small benchmarks on i7-12800H, 32GB RAM. 41044 tests on 31 devices**:
PR: `anta nrfu json --output anta_results.json  78.45s user 4.17s system 82% cpu 1:40.15 total`

main: `anta nrfu json --output anta_results.json  74.72s user 3.85s system 75% cpu 1:43.48 total`

**With max concurrency of 25000 (instead of default 50000):**
`anta nrfu json --output anta_results.json  76.49s user 4.14s system 66% cpu 2:01.39 total`

As expected, with a limit, it takes more time but less CPU usage.

# Checklist:

- [x] Update FAQ
- [x] Update benchmark + unit tests
- [x] #826 
- [x] Confirm that the number of open file descriptors never exceed device * max_connections (100 by default)


anta/runner.py
```diff
@@ -75,6 +76,57 @@ def log_cache_statistics(devices: list[AntaDevice]) -> None:
...
 
+async def run_tests(tests_generator: AsyncGenerator[Coroutine[Any, Any, TestResult], None], limit: int) -> AsyncGenerator[TestResult, None]:
+    """Run tests with a concurrency limit.
+
+    This function takes an asynchronous generator of test coroutines and runs them
+    with a limit on the number of concurrent tests. It yields test results as each
+    test completes.
+
+    Args:
+    ----
```
gmuloc, line 87:
```suggestion
    Parameters
    -----------
```

anta/runner.py
```diff
@@ -157,23 +209,24 @@ def prepare_tests(
...
 
-def get_coroutines(selected_tests: defaultdict[AntaDevice, set[AntaTestDefinition]]) -> list[Coroutine[Any, Any, TestResult]]:
-    """Get the coroutines for the ANTA run.
+async def generate_tests(selected_tests: defaultdict[AntaDevice, set[AntaTestDefinition]]) -> AsyncGenerator[Coroutine[Any, Any, TestResult], None]:
+    """Generate the coroutines for the ANTA run.
+
+    It creates an async generator of coroutines which are created by the `test` method of the AntaTest instances. Each coroutine is a test to run.
 
     Args:
     ----
```
gmuloc, line 218:
```suggestion
    Parameters
    -----------
```

anta/runner.py
```diff
@@ -77,6 +104,61 @@ def log_cache_statistics(devices: list[AntaDevice]) -> None:
...
+    tests_generator
+        An asynchronous generator that yields test coroutines.
+    limit
+        The maximum number of concurrent tests to run.
+
+    Yields
+    ------
+        The result of each completed test.
+    """
+    # NOTE: The `aiter` built-in function is not available in Python 3.9
```
gmuloc, line 127: so this will fail in 3.9?

carl-baillargeon: `aiter()` and `anext()` were released in 3.10: https://docs.python.org/3/library/functions.html#aiter so we need to use the dunders instead (since we support 3.9 RIP)

gmuloc: not RIP yet: https://devguide.python.org/versions/

gmuloc: ok so this can be like this until we drop 3.9 understood thanks!

anta/device.py
```diff
@@ -33,6 +34,78 @@
...
+ANTA_DEFAULT_TIMEOUT = 30.0
+
+
+def get_httpx_limits() -> Limits:
+    """Adjust the underlying HTTPX client resource limits.
+
+    The limits are set using the following environment variables:
+        - ANTA_MAX_CONNECTIONS: Maximum number of allowable connections.
+        - ANTA_MAX_KEEPALIVE_CONNECTIONS: Number of allowable keep-alive connections.
+        - ANTA_KEEPALIVE_EXPIRY: Time limit on idle keep-alive connections in seconds.
```
gmuloc, line 46: Link to httpx doc

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+!!! note "Default Values"
+    The values specified in the following examples are the actual default values used by ANTA. See the [Performance Tuning](#performance-tuning) section for guidance on adjusting these values.
+
+1. **Test Concurrency**:
+
+    ```bash
+    export ANTA_MAX_CONCURRENCY=10000
+    ```
+
+    This limits the number of concurrent test coroutines to prevent event loop overload. ANTA will schedule coroutines up to this limit, then wait for some to complete before scheduling more.
```
gmuloc, line 82:
```suggestion
    ANTA limits the number of concurrent test coroutines to prevent event loop overload. The runner schedules coroutines up to this limit, then waits for some to complete before scheduling more. The following variable can be used to tune this limit:
```

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+2. **File Descriptors**:
+
+    ```bash
+    export ANTA_NOFILE=16384
+    ```
+
+    Sets the maximum number of open file descriptors for the ANTA process, usually for handling device connections. At startup, ANTA sets its process’s soft limit to the maximum allowed (up to `16384`). This adjustment is necessary because the soft limit is typically set to `1024`, while the hard limit is often higher (system-dependent). If ANTA’s hard limit is lower than the number of selected tests in ANTA, the process may request more file descriptors than the operating system allows, causing an error. In such cases, a WARNING is displayed at startup.
```
gmuloc, line 90: Same I would first described ANTA behabior and then states that this variable can be used

gmuloc, line 90:
```suggestion
    Sets the maximum number of open file descriptors for the ANTA process, usually for handling device connections. At startup, ANTA sets its process soft limit to the maximum allowed (up to `16384`). This adjustment is necessary because the soft limit is typically set to `1024`, while the hard limit is often higher (system-dependent). If ANTA hard limit is lower than the number of selected tests in ANTA, the process may request more file descriptors than the operating system allows, causing an error. In such cases, a WARNING is displayed at startup.
```

Also add a note that it does not work on Windows system today.
Maybve we can show the warning

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+    To address this, consider increasing the hard limit for the user starting the ANTA process. You can check the current hard limit for a user by running the command `ulimit -n -H` in the terminal. To set a new limit, create the file `/etc/security/limits.d/10-anta.conf` with the following content:
```
gmuloc, line 92: only for UNIX systems add a note that this is system dependent

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+    ```bash
+    <user> hard nofile <value>
+    ```
+
+    Replace `<user>` with the username used to start the ANTA process, and `<value>` with the desired hard limit. The maximum value will depend on your system. After creating this file, log out of your current session and log back in for the changes to take effect.
+
+### Connection Management
+
+ANTA uses the `httpx` library as its underlying HTTP client for device connections. Each device has his own connection pool with a default maximum of `100` connections. On large fabrics, this can lead to a very high number of connections (n x 100), which may overwhelm the ANTA process. Connection pooling can be tuned via:
```
gmuloc, line 102: specify for `AsyncEapiDevice`

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+export ANTA_MAX_CONNECTIONS=100
+
+# Number of allowable keep-alive connections
+export ANTA_MAX_KEEPALIVE_CONNECTIONS=20
+
+# Time limit on idle keep-alive connections in seconds
+export ANTA_KEEPALIVE_EXPIRY=5.0
+```
+
+**Best Practice**: For large fabrics, limiting connections (5-10 per device) has shown optimal performance in testing. `ANTA_MAX_KEEPALIVE_CONNECTIONS` should be lower than `ANTA_MAX_CONNECTIONS`. See [HTTPX documentation](https://www.python-httpx.org/advanced/resource-limits/) for more details.
```
gmuloc, line 115: make this a box

```
!!! tip "Best Practices"
   ..
```

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+export ANTA_READ_TIMEOUT=30.0
+
+# Maximum duration to wait for a chunk of data to be sent (for example, a chunk of the request body)
+export ANTA_WRITE_TIMEOUT=30.0
+
+# Maximum duration to wait for acquiring a connection from the connection pool
+export ANTA_POOL_TIMEOUT=30.0
+```
+
+If not set, these values default to the global `timeout` of ANTA, which is `30.0` seconds by default. The global timeout can also be set via the `--timeout` command-line option. See [HTTPX documentation](https://www.python-httpx.org/advanced/timeouts/) for more details.
```
gmuloc, line 138: Same would put the text first with ANTA behavior and then explains how it can be finetuned and link to HTTPX

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+For optimal single-process performance, consider the following tuning parameters:
+
+1. Adjust concurrency based on test count and system resources:
+
+    ```bash
+    # For 250 devices with 100 tests each (25,000 total tests)
+    export ANTA_MAX_CONCURRENCY=25000
+    ```
+
+    Even though the default value is `10000`, it may not be optimal for your system. You can increase or decrease this value gradually to find the best setting for your environment.
```
gmuloc, line 151: hopw do I determine what is best?runtime?

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+    These values are a good starting point for large fabrics. Adjust them based on your fabric size and performance testing.
+
+3. Increase timeouts for stability:
+
+    ```bash
+    export ANTA_TIMEOUT=3600.0
+    ```
+
+    This is **very** important for large fabrics. Increase the global timeout to prevent test failures due to timeouts. You shouldn't need to adjust the other timeout values unless you have specific requirements.
```
gmuloc, line 168: Hwo does this manifest?
I feel very important should be all bold and potentially this should be number 1 in the list :)

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+## 🔀 Multi-Process Scaling
+
+When single-process optimization isn't enough, you can scale ANTA horizontally by running multiple instances in parallel. This approach is particularly useful for large fabrics with 500+ devices. This section will provide a few examples of how to achieve this using the common tool [`GNU Parallel`](https://www.gnu.org/software/parallel/).
```
gmuloc, line 172:
```suggestion
When single-process optimization isn't enough, you can scale ANTA horizontally by running multiple instances in parallel. This approach is particularly useful for large fabrics with 500+ devices. This section will provide a few examples of how to achieve this using the common tool [`GNU Parallel`](https://www.gnu.org/software/parallel/).
```
```suggestion
When single-process optimization isn't enough, you can scale ANTA horizontally by running multiple instances in parallel. This approach is particularly useful for large fabrics with 500+ devices. This section provides a couple of examples of how to achieve this using the common tool [`GNU Parallel`](https://www.gnu.org/software/parallel/).
```

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+By leveraging `GNU Parallel` and proper test organization, we can significantly reduce execution time and improve overall performance while maintaining accurate test results.
```
gmuloc, line 174:
```suggestion
By leveraging `GNU Parallel` and proper test organization, execution time can be significantly reduced, improving overall performance while maintaining accurate test results.
```

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+- 🎯 Identify natural test boundaries (e.g., devices, PODs, roles, etc.)
+- 📋 Plan your test catalog(s) accordingly
+- 💻 Understand basic Linux command-line operations
+
+### Installation Requirements
+
+!!! info "System Requirements"
+    The following procedures and examples were tested on **Ubuntu 22.04**, using `GNU Parallel` version `20241022` and `yq` version `v4.44.3`. Please refer the tool's documentation for specific installation instructions.
+
+#### Installing Required Tools
```
gmuloc, line 190: title is not needed as there is no other sub title (or maybe should be called `Ubuntu`

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+    (.dev) ~ yq --version
+    yq (https://github.com/mikefarah/yq/) version v4.44.3
+    ```
+
+### Scaling Strategies
+
+#### Device-Based Parallelization
+
+In this approach, we run ANTA instances in parallel, with each instance handling a specific device. This strategy is effective when:
```
gmuloc, line 236:
```suggestion
In this approach, instances of ANTA targeting a single device each are run in parallel. This strategy is effective when:
```

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+# DC1-SPINE1.yml
+anta.tests.connectivity:
+- VerifyLLDPNeighbors:
+    neighbors:
+    - neighbor_device: DC1-LEAF1A
+      neighbor_port: Ethernet2
+      port: Ethernet1
+    result_overwrite:
+      custom_field: 'Local: Ethernet1 - Remote: DC1-LEAF1A Ethernet2'
+```
```
gmuloc, line 276: Maybe lets use tabs or columns ?
````suggestion
```yaml
# DC1-SPINE2.yml
anta.tests.connectivity:
- VerifyLLDPNeighbors:
    neighbors:
    - neighbor_device: DC1-LEAF1A
      neighbor_port: Ethernet1
      port: Ethernet1
    result_overwrite:
      custom_field: 'Local: Ethernet1 - Remote: DC1-LEAF1A Ethernet1'
```

```yaml
# DC1-SPINE1.yml
anta.tests.connectivity:
- VerifyLLDPNeighbors:
    neighbors:
    - neighbor_device: DC1-LEAF1A
      neighbor_port: Ethernet2
      port: Ethernet1
    result_overwrite:
      custom_field: 'Local: Ethernet1 - Remote: DC1-LEAF1A Ethernet2'
```
````

File: docs/advanced_usages/scaling.md [hunk 18]
```diff
@@ -0,0 +1,635 @@
...
+
+anta.tests.routing:
+  generic:
+    - VerifyRoutingProtocolModel:
+        model: multi-agent
+        filters:
+          tags: ["SPINE"]
+```
+
+#### Optimizing Catalog Distribution
```
gmuloc, line 337: Should probably come first in the section?

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+Benefits:
+
+- Reduced memory usage per ANTA instance
+- Faster catalog processing
+- More efficient test execution
+
+### Implementation
+
+#### Device-Based Implementation
```
gmuloc, line 353: I would move Implementation under each stratgey in a subtitle for easier navigation

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+4. Run the script:
+
+    ```bash
+    ./run-tag.sh
+    ```
+
+5. The script will execute ANTA tests for each tag in parallel, generating a JSON report for each tag in the `nrfu_reports` directory.
+
+### Performance Considerations
```
gmuloc, line 481: same this secrtion shoul probably come before Scaling strategies

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+     - Example: For 8 cores, use `-j7`
+
+2. **Memory requirements**:
+
+     - Each ANTA instance requires memory
+     - Monitor system resources during execution using `htop` or similar tools
+
+#### JSON vs YAML
+
+When working with large catalogs and inventories, using JSON format instead of YAML can provide better performance as ANTA loads JSON files more efficiently. If you choose to use JSON:
```
gmuloc, line 499: I think thats fair?
```suggestion
When working with large catalogs and inventories, using JSON format instead of YAML can provide better performance as Python loads JSON files more efficiently. If you choose to use JSON:
```

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+  Example of converting YAML to JSON:
+
+  ```bash
+  yq -o=json eval 'anta_inventory.yaml' > anta_inventory.json
+  yq -o=json eval 'anta_catalog.yaml' > anta_catalog.json
+  ```
+
+### Results Management
+
+ANTA can output results in various formats:
```
gmuloc, line 521: do you need to repeat these? I would just point to the NRFU page documentatin and say you use JSON (less maintenance)

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+    Please refer to the [ANTA CLI documentation](https://anta.arista.com/stable/cli/nrfu/) for more information on result output formats.
+
+    !!! note "Output Formats"
+        The scripts provided in this guide output JSON files for each device or tag. You can modify them to output in other formats as needed.
+
+3. Optional: You can merge the JSON results and generate a JUnit report using the following Python script (requires `junitparser`):
+
+    ```python
+    from pathlib import Path
```
gmuloc, line 543: please use snippets for this and we will try to test it :)

tests/benchmark/test_runner.py
```diff
@@ -7,8 +7,7 @@
 
 from typing import TYPE_CHECKING
 
-from anta.result_manager import ResultManager
-from anta.runner import get_coroutines, prepare_tests
+from anta.runner import prepare_tests
```
gmuloc, line 10: we can add back get_coroutines for now

gmuloc (commented):
Lots on comment on the guide but a very big **Thank You** for this amazing guide :) 

🎉 🎈

anta/settings.py
```diff
@@ -0,0 +1,195 @@
...
+        keepalive_expiry=settings.keepalive_expiry,
+    )
+
+
+def get_httpx_timeout(default_timeout: float | None) -> Timeout:
+    """Get the HTTPX Timeout object from environment variables.
+
+    Parameters
+    ----------
+    default_timeout : float | None
```
mtache, line 148: Default values are defined twice here.
Once in this file, the other definition is in the timeout CLI option.
I would suggest to keep the default values in this file, remove the default value in Click and point to the documentation.

anta/cli/utils.py
```diff
@@ -163,6 +190,7 @@ def core_options(f: Callable[..., Any]) -> Callable[..., Any]:
         show_envvar=True,
         envvar="ANTA_TIMEOUT",
         show_default=True,
+        type=FLOAT_OR_NONE,
```
mtache, line 193: I would suggest to remove the default value here since it is defined in the anta.settings module already.
Then you can also get rid of the FLOAT_OR_NONE type :)

mkdocs.yml
```diff
@@ -181,9 +181,10 @@ nav:
     - Debug commands: cli/debug.md
     - Tag Management: cli/tag-management.md
   - Advanced Usages:
-    - Caching in ANTA: advanced_usages/caching.md
     - Developing ANTA tests: advanced_usages/custom-tests.md
     - ANTA as a Python Library: advanced_usages/as-python-lib.md
+    - Caching in ANTA: advanced_usages/caching.md
+    - Scaling ANTA: advanced_usages/scaling.md
```
mtache, line 187: Can we add a Settings section that render the BaseSettings models in anta.settings?

anta/settings.py
```diff
@@ -0,0 +1,195 @@
...
+# Default value for the maximum number of open file descriptors for the ANTA process
+DEFAULT_NOFILE = 16384
+
+
+class FileDescriptorSettings(BaseSettings):
+    """Environment variable for configuring the maximum number of open file descriptors for the ANTA process.
+
+    On POSIX systems, this value is used to set the soft limit for the current process.
+    The value cannot exceed the system's hard limit.
+    """
```
mtache, line 41: Missing ANTA_NOFILE attribute in the doctsring.

anta/settings.py
```diff
@@ -0,0 +1,195 @@
...
+
+    model_config = SettingsConfigDict(env_prefix="ANTA_")
+
+    nofile: PositiveInt = Field(default=DEFAULT_NOFILE)
+
+
+class MaxConcurrencySettings(BaseSettings):
+    """Environment variable for configuring the maximum number of concurrent tests in the event loop."""
```
mtache, line 49: Missing ANTA_MAX_CONCURRENCY attribute in the docstring

anta/settings.py
```diff
@@ -0,0 +1,195 @@
...
+
+    model_config = SettingsConfigDict(env_prefix="ANTA_")
+
+    max_concurrency: PositiveInt = Field(default=DEFAULT_MAX_CONCURRENCY)
+
+
+class HttpxResourceLimitsSettings(BaseSettings):
+    """Environment variables for configuring the underlying HTTPX client resource limits.
+
+    The limits are set using the following environment variables:
```
mtache, line 59: Use attribute section to have a table in mkdocs?

anta/_runner.py
```diff
@@ -0,0 +1,508 @@
...
+                    tests_ended = True
+                    logger.debug("All tests have been added to the pending set.")
+                else:
+                    # Ensure the coroutine is scheduled to run and add it to the pending set
+                    tests_pending.add(asyncio.create_task(test))
+                    logger.debug("Added a test to the pending set: %s", test)
+
+            if len(tests_pending) >= self.max_concurrency:
+                logger.debug("Concurrency limit reached: %s tests running. Waiting for tests to complete.", self.max_concurrency)
+
```
mtache, line 480:
```suggestion
        while tests_pending or not tests_ended:
            async for test in tests:
                # Add tests to the pending set until the limit is reached or no more tests are available
                if len(tests_pending) < self.max_concurrency:
                    # Ensure the coroutine is scheduled to run and add it to the pending set
                    tests_pending.add(asyncio.create_task(test))
                    logger.debug("Added a test to the pending set: %s", test)
                else:
                    logger.debug("Concurrency limit reached: %s tests running. Waiting for tests to complete.", self.max_concurrency)
                    break
            if len(tests_pending) < self.max_concurrency and not tests_ended:
                tests_ended = True
                logger.debug("All tests have been added to the pending set.")

```

anta/_runner.py
```diff
@@ -0,0 +1,508 @@
...
+            if not tests_pending:
+                logger.debug("No pending tests and all tests have been processed. Exiting.")
+                return
```
mtache, line 483:
```suggestion
        if self.max_concurrency <= 0:
            msg = "Concurrency limit must be greater than 0."
            raise RuntimeError(msg)

        tests_ended = False
        tests_pending: set[Task[TestResult]] = set()

        while tests_pending or not tests_ended:
            async for test in generator:
                # Add tests to the pending set until the limit is reached or no more tests are available
                if len(tests_pending) < self.max_concurrency:
                    # Ensure the coroutine is scheduled to run and add it to the pending set
                    tests_pending.add(asyncio.create_task(test))
                    logger.debug("Added a test to the pending set: %s", test)
                else:
                    logger.debug("Concurrency limit reached: %s tests running. Waiting for tests to complete.", self.max_concurrency)
                    break
            if len(tests_pending) < self.max_concurrency and not tests_ended:
                tests_ended = True
                logger.debug("All tests have been added to the pending set.")

            if not tests_pending:
                logger.debug("No pending tests and all tests have been processed. Exiting.")
                return
```

anta/device.py
```diff
@@ -378,15 +388,31 @@ def __init__(  # noqa: PLR0913
...
+            "username": username,
+            "password": password,
+            "proto": proto,
+            "timeout": httpx_timeout if httpx_timeout is not None else timeout,
+        }
+        if httpx_limits is not None:
+            session_settings["limits"] = httpx_limits
+            self.max_connections = httpx_limits.max_connections
+        else:
+            self.max_connections = DEFAULT_HTTPX_MAX_CONNECTIONS
```
mtache, line 408: More robust logic to instantiate the `asynceapi.Device` object then retrieve `self._session.limits.max_connection`.
Because of https://github.com/encode/httpx/blob/10b7295922741b91a15751029e6ad3e8e5efb9f3/httpx/_client.py#L655, httpx default values will be used.

anta/device.py
```diff
@@ -354,14 +358,20 @@ def __init__(  # noqa: PLR0913
         tags
             Tags for this device.
         timeout
-            Timeout value in seconds for outgoing API calls.
-        insecure
-            Disable SSH Host Key validation.
+            Global timeout value in seconds for outgoing eAPI calls.
+        httpx_timeout
+            Optional HTTPX Timeout object for fine-grained timeout configuration.
+            If provided, it will override the global timeout.
```
mtache, line 364: timeout defined twice.
Use `timeout: httpx.Timeout` object instead and remove `timeout: float`.
Use a read property to get the `Timeout` instance of this `AyncEOSDevice` instance.

anta/device.py
```diff
@@ -378,15 +388,31 @@ def __init__(  # noqa: PLR0913
...
+        ssh_settings = {"host": host, "port": ssh_port, "username": username, "password": password, "client_keys": CLIENT_KEYS}
         if insecure:
-            ssh_params["known_hosts"] = None
-        self._ssh_opts: SSHClientConnectionOptions = SSHClientConnectionOptions(
-            host=host, port=ssh_port, username=username, password=password, client_keys=CLIENT_KEYS, **ssh_params
-        )
+            ssh_settings["known_hosts"] = None
+        self._ssh_opts: SSHClientConnectionOptions = SSHClientConnectionOptions(**ssh_settings)
 
     def __rich_repr__(self) -> Iterator[tuple[str, Any]]:
```
mtache, line 417: implement `timeout` and `limits` ?

anta/settings.py
```diff
@@ -0,0 +1,195 @@
...
+# Default values for HTTPX resource limits
+DEFAULT_HTTPX_MAX_CONNECTIONS = 100
+DEFAULT_HTTPX_MAX_KEEPALIVE_CONNECTIONS = 20
+DEFAULT_HTTPX_KEEPALIVE_EXPIRY = 5.0
+
+# Default values for HTTPX timeouts
+DEFAULT_HTTPX_CONNECT_TIMEOUT = 5.0
+DEFAULT_HTTPX_READ_TIMEOUT = 5.0
+DEFAULT_HTTPX_WRITE_TIMEOUT = 5.0
+DEFAULT_HTTPX_POOL_TIMEOUT = 5.0
```
mtache, line 27: Can we use https://github.com/encode/httpx/blob/10b7295922741b91a15751029e6ad3e8e5efb9f3/httpx/_config.py#L246 instead of copying the default value?
Not sure if the value will render in mkdocs.

tests/benchmark/test_anta.py
```diff
@@ -37,6 +37,7 @@ def test_anta_dry_run(
 
     results = session_results[request.node.callspec.id]
 
+    # TODO: Use AntaRunner in ANTA v2.0.0
```
mtache, line 40: Is this needed?
I think the `anta.runner.main` function uses the `AntaRunner` class already.

tests/benchmark/test_anta.py
```diff
@@ -69,6 +70,7 @@ def test_anta(
...
```
mtache, line 73: Same here.

mtache (changes requested):
Default value code could be simplified/cleaned.
Good idea to have this `anta.settings` module, we should use it for `ANTA_DEBUG`, in another PR ;)

anta/settings.py
gmuloc: if it changes upstream it may make us sad one day

anta/settings.py
carl-baillargeon: Agreed with removing the default value in the CLI.

anta/cli/utils.py
carl-baillargeon: Same as above, agreed with removing the default value here but we do need to keep FLOAT_OR_NONE to support None (no timeout) at the CLI. We do not support None today:
```
(.venv) ~/git_projects/anta (main ✔) anta nrfu --timeout None table
Usage: anta nrfu [OPTIONS] COMMAND [ARGS]...
Try 'anta nrfu --help' for help.

Error: Invalid value for '--timeout': 'None' is not a valid float.
```

anta/_runner.py
carl-baillargeon: It looks equivalent right?

anta/device.py
carl-baillargeon: I believe you are suggesting to not store `max_connections` as a separate attribute, but rather create a read-only property that retrieves it directly from HTTPX? `Limits` is injected down the line to create the httpcore connection pool. We could potentially retrieve it by doing `self._session._transport._pool._max_connections` but since these are all privates it's too risky IMO.

anta/device.py
carl-baillargeon: Ah yes I didn't know HTTPX supported different types. Type would be `timeout: Timeout | float | None` to avoid breaking change. Should we also support the tuple type (below)?
```
TimeoutTypes = Union[
    Optional[float],
    Tuple[Optional[float], Optional[float], Optional[float], Optional[float]],
    "Timeout",
]
```

carl-baillargeon: I think we should also replace `httpx_limits` with just `limits` to be consistent with `timeout` then.

tests/benchmark/test_anta.py
carl-baillargeon: We can consider removing `main()` altogether and use `AntaRunner` everywhere in v2.0.0.

tests/benchmark/test_anta.py
carl-baillargeon: I put this TODO everywhere we are using `main()`.

anta/settings.py
carl-baillargeon: We could put our own default values here instead with the current 30s default we have. From our testing we have a pretty good idea what works best:
```python
# Default values for HTTPX resource limits
DEFAULT_HTTPX_MAX_CONNECTIONS = 10
DEFAULT_HTTPX_MAX_KEEPALIVE_CONNECTIONS = 10
DEFAULT_HTTPX_KEEPALIVE_EXPIRY = 5.0

# Default values for HTTPX timeouts
DEFAULT_HTTPX_CONNECT_TIMEOUT = 30.0
DEFAULT_HTTPX_READ_TIMEOUT = 30.0
DEFAULT_HTTPX_WRITE_TIMEOUT = 30.0
DEFAULT_HTTPX_POOL_TIMEOUT = None
```

carl-baillargeon: I decided to put our own default values since they were tested. d8ca4acfe433b8a73a82a53cca7f2cf51cd177ad

anta/device.py
carl-baillargeon: `timeout` is already in `_session`. Added `max_connections` in d8ca4acfe433b8a73a82a53cca7f2cf51cd177ad.

anta/device.py
carl-baillargeon: Fixed in d8ca4acfe433b8a73a82a53cca7f2cf51cd177ad.

anta/device.py
carl-baillargeon: Added in d8ca4acfe433b8a73a82a53cca7f2cf51cd177ad around a try-expect block if not available.

anta/cli/utils.py
carl-baillargeon: Added `float` type to support `inf` for no timeout in d8ca4acfe433b8a73a82a53cca7f2cf51cd177ad.

anta/settings.py
carl-baillargeon: Fixed in d8ca4acfe433b8a73a82a53cca7f2cf51cd177ad.

anta/_runner.py
carl-baillargeon: I didn't troubleshoot much but coroutines are not awaited properly with this code:
```
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
/home/cbaillar/git_projects/anta/anta/_runner.py:449: RuntimeWarning: coroutine 'VerifySSHStatus.test' was never awaited
  async for test in generator:
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
/home/cbaillar/git_projects/anta/anta/_runner.py:449: RuntimeWarning: coroutine 'VerifyRoutingProtocolModel.test' was never awaited
  async for test in generator:
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
/home/cbaillar/git_projects/anta/anta/_runner.py:449: RuntimeWarning: coroutine 'VerifyAcctConsoleMethods.test' was never awaited
  async for test in generator:
  ```

File: anta/settings.py (hunk 28 above)
Comment (carl-baillargeon) [reply anta/settings.py:49]: Fixed in dee3bc9c479496fb5f9dac84d9e1e751b5dede0c

File: mkdocs.yml (hunk 27 above)
Comment (carl-baillargeon) [reply mkdocs.yml:191]: I added it to the API doc in dee3bc9c479496fb5f9dac84d9e1e751b5dede0c.

File: anta/settings.py (hunk 29 above)
Comment (carl-baillargeon) [reply anta/settings.py]: Fixed in dee3bc9c479496fb5f9dac84d9e1e751b5dede0c.

File: anta/settings.py (hunk 30 above)
Comment (carl-baillargeon) [reply anta/settings.py]: Fixed in dee3bc9c479496fb5f9dac84d9e1e751b5dede0c.

File: docs/advanced_usages/scaling.md [hunk 39]
```diff
@@ -0,0 +1,224 @@
...
+
+Remember to:
+
+- ✅ Optimize resource usage through environment variables
+- 📋 Consider file format performance impacts
+- 📊 Choose appropriate output formats for your needs
+- ⚡ Monitor and adjust based on your specific environment
+
+!!! tip "Need Help?"
+    Performance tuning can be complex and highly dependent on your specific deployment. See the warning "Advanced Guide" at the beginning of this document for assistance options.
```
gmuloc, line 198: can we put an anchor?

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+# 📖 Introduction
+
+!!! warning "Advanced Guide"
+    This document covers advanced ANTA configuration and scaling strategies. If you need assistance with specific deployments or optimizations:
+
+    - 📧 Contact your favorite Arista Systems Engineer
+    - 💻 Open an issue or a discussion on ANTA [GitHub repository](https://github.com/aristanetworks/anta)
+    - 🤝 Join the discussion in our community channels
+
+    This guide also assumes you have a basic understanding of ANTA and are familiar with its core concepts. If you're new to ANTA, we recommend starting with the [ANTA Getting Started](../getting-started.md) before proceeding.
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+
+!!! note "Key Concept"
+    While `asyncio` provides excellent concurrency, there are practical limits to how many coroutines can be efficiently managed in a single event loop running on one CPU core. High coroutine counts increase event loop overhead, lead to frequent context switching, consume more memory, and may degrade performance.
+
+## Resource Management
+
+This section describes the environment variables to control resource usage provided by ANTA.
+
+!!! tip "Default Values"
+    The values specified in the following examples are the actual default values used by ANTA. See the [Performance Tuning](#performance-tuning) section for guidance on adjusting these values.
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+
+ANTA limits the number of concurrent test coroutines to prevent event loop overload. The runner schedules coroutines up to this limit, then waits for some to complete before scheduling more. The following variable can be used to tune this limit:
+
+```bash
+export ANTA_MAX_CONCURRENCY=10000
+```
+
+### **File Descriptors**
+
+ANTA manages the number of open file descriptors, which are primarily used for handling device connections. At startup, on POSIX systems (Linux/macOS), ANTA attempts to set its process's soft limit to the maximum allowed value (up to 16384). This adjustment is necessary because the default soft limit (typically 1024) may be too low for large-scale testing, while the system's hard limit is often higher. If the system's hard limit is lower than the number of file descriptors ANTA needs, you'll see a warning at startup. The following environment variable controls this setting:
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+
+For POSIX systems, if you need to increase the hard limit:
+
+- Check your current hard limit:
+
+```bash
+ulimit -n -H
+```
+
+- Set a new limit by creating `/etc/security/limits.d/10-anta.conf`:
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+
+```bash
+export ANTA_NOFILE=16384
+```
+
+On Windows (non-POSIX systems), ANTA cannot adjust file descriptor limits. In these cases, ANTA will:
+
+- Use Python `sys.maxsize` as the limit
+- Display a warning message: "Running on a non-POSIX system, cannot adjust the maximum number of file descriptors"
+- Continue operation with system defaults
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+<user> hard nofile <value>
+```
+
+Replace `<user>` with the username running ANTA, and `<value>` with your desired limit (system-dependent).
+
+- Log out and back in for changes to take effect.
+
+## Connection Management
+
+For [`AsyncEosDevice`](../api/device.md#anta.device.AsyncEOSDevice) implementations (default in ANTA), ANTA uses the [`httpx`](https://www.python-httpx.org/) library as its underlying HTTP client for device connections. Each device has his own connection pool and HTTPX uses a default maximum of `100` connections. On large fabrics, this can lead to a very high number of connections (n x 100), which may overwhelm the ANTA process. Note that custom `AntaDevice` subclasses might implement different HTTP settings. For `AsyncEosDevice`, connection pooling can be tuned via the following environment variables. See [HTTPX documentation](https://www.python-httpx.org/advanced/resource-limits/) for more details.
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+export ANTA_MAX_CONNECTIONS=10
+
+# Number of allowable keep-alive connections
+export ANTA_MAX_KEEPALIVE_CONNECTIONS=10
+
+# Time limit on idle keep-alive connections in seconds
+export ANTA_KEEPALIVE_EXPIRY=5.0
+```
+
+!!! warning "Variable relationships"
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+
+# Maximum duration to wait for a chunk of data to be sent (for example, a chunk of the request body)
+export ANTA_WRITE_TIMEOUT=30.0
+
+# Maximum duration to wait for acquiring a connection from the connection pool
+export ANTA_POOL_TIMEOUT=inf
+```
+
+!!! warning "Global timeout"
+    Setting a global timeout via the `--timeout` command-line option or the `ANTA_TIMEOUT` environment variable will override **ALL** individual timeout settings.
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+    export ANTA_POOL_TIMEOUT=inf
+    ```
+
+    This is **very** important for large fabrics. Setting appropriate timeouts for each operation type prevents test failures while maintaining proper error detection. The `ANTA_POOL_TIMEOUT=inf` default setting is crucial for large-scale deployments because:
+
+      - ANTA launches multiple test coroutines simultaneously (up to `ANTA_MAX_CONCURRENCY`)
+      - With thousands of tests, not all can acquire a connection immediately
+      - The connection pool queues requests when all connections are in use
+      - Without `ANTA_POOL_TIMEOUT=inf`, requests might fail if they can't get a connection within the specified timeout
+      - Setting it to `None` allows the connection pool to manage request queuing naturally
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+
+2. Adjust concurrency based on test count and system resources:
+
+    ```bash
+    # For 250 devices with 100 tests each (25,000 total tests)
+    export ANTA_MAX_CONCURRENCY=25000
+    ```
+
+    Even though the default value is `10000`, it may not be optimal for your system. The optimal value depends on your CPU usage, memory consumption, and total execution time. Start with the default and adjust in increments while monitoring these metrics to find the best balance for your environment.
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+      - ANTA launches multiple test coroutines simultaneously (up to `ANTA_MAX_CONCURRENCY`)
+      - With thousands of tests, not all can acquire a connection immediately
+      - The connection pool queues requests when all connections are in use
+      - Without `ANTA_POOL_TIMEOUT=inf`, requests might fail if they can't get a connection within the specified timeout
+      - Setting it to `None` allows the connection pool to manage request queuing naturally
+
+2. Adjust concurrency based on test count and system resources:
+
+    ```bash
+    # For 250 devices with 100 tests each (25,000 total tests)
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+
+    The default values are a good starting point for large fabrics. Adjust them based on your fabric size and performance testing.
+
+4. Consider JSON over YAML for better performance:
+
+    - JSON files load more efficiently than YAML in Python
+    - Convert large YAML files to JSON (you can use tools like `jq` or `yq`)
+
+        ```bash
+        yq -o=json eval 'inventory.yaml' > inventory.json
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
carl-baillargeon: Removed the guide. Check FAQ instead.

docs/advanced_usages/scaling.md
```diff
@@ -0,0 +1,224 @@
...
+!!! warning "Advanced Guide"
+    This document covers advanced ANTA configuration and scaling strategies. If you need assistance with specific deployments or optimizations:
+
+    - 📧 Contact your favorite Arista Systems Engineer
+    - 💻 Open an issue or a discussion on ANTA [GitHub repository](https://github.com/aristanetworks/anta)
+    - 🤝 Join the discussion in our community channels
+
+    This guide also assumes you have a basic understanding of ANTA and are familiar with its core concepts. If you're new to ANTA, we recommend starting with the [ANTA Getting Started](../getting-started.md) before proceeding.
+
+**ANTA** (Arista Network Test Automation) is continually evolving to meet the demands of large-scale network validation. As networks grow in size and complexity, the need for efficient, scalable testing becomes increasingly important. This guide explores comprehensive strategies for scaling ANTA to handle large network fabrics effectively.
```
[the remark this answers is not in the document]
carl-baillargeon: Removed the guide. Check FAQ instead.

anta/settings.py
```diff
@@ -0,0 +1,119 @@
...
+    ROUND_ROBIN = "round-robin"
+    DEVICE_BY_DEVICE = "device-by-device"
+    DEVICE_BY_COUNT = "device-by-count"
+
+    def __str__(self) -> str:
+        """Override the __str__ method to return the value of the Enum, mimicking the behavior of StrEnum."""
+        return self.value
+
+
+class AntaRunnerSettings(BaseSettings):
```
gmuloc, line 51: not for this PR but probably we should think about making a config file down the line (though I hate interaction between config file/ CLI / env variable :p)

anta/tools.py
```diff
@@ -415,3 +418,45 @@ def format_data(data: dict[str, bool]) -> str:
...
+
+    coros_ended = False
+    pending: set[asyncio.Task[T]] = set()
+
+    while pending or not coros_ended:
+        while len(pending) < limit and not coros_ended:
+            try:
+                # NOTE: The `anext` built-in function is not available in Python 3.9
+                coro = await coroutines.__anext__()  # pylint: disable=unnecessary-dunder-call
+            except StopAsyncIteration:  # noqa: PERF203
```
gmuloc, line 451: lets add a comment that we don't want to stop the loop in case an exception is raised and thats why we ignore PERF203
```suggestion
            except StopAsyncIteration:  # noqa: PERF203
```

tests/units/test_tools.py
```diff
@@ -527,3 +534,78 @@ def test_convert_categories(test_input: list[str], expected_raise: AbstractConte
...
+    async def _mock_test_coro(self, result: Result) -> Result:
+        """Mock coroutine simulating a test."""
+        # Simulate some work
+        await asyncio.sleep(0.1)
+        return result
+
+    async def _create_test_generator(self, results: Sequence[Result]) -> AsyncGenerator[Coroutine[Any, Any, Result], None]:
+        """Create a test generator yielding mock test coroutines."""
+        for result in results:
+            yield self._mock_test_coro(result)
```
gmuloc, line 563: should be a pytest fixture it seems

anta/_runner.py
gmuloc: TODO for ANTA v1.4.0 - need to understand why `async for` + `break` is _sometimes_ raising this kind of errors
```
 RuntimeWarning: coroutine 'VerifyAcctConsoleMethods.test' was never awaited
  async for test in generator:
```

we want to use `async for`

anta/_runner.py
```diff
@@ -0,0 +1,473 @@
...
+@dataclass(frozen=True)
+class AntaRunnerInventoryStats:
+    """Store inventory filtering statistics of an ANTA run.
+
+    This class maintains counters for tracking how the device inventory is filtered
+    during test setup.
+
+    Attributes
+    ----------
+    total : int
```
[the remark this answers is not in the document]
carl-baillargeon: As discussed, keeping type hints for IDE's.

anta/device.py
```diff
@@ -385,15 +392,26 @@ def __init__(  # noqa: PLR0913
...
+        # Build the session settings for the `asynceapi` client
+        session_settings: dict[str, Any] = {
+            "host": host,
+            "port": port,
+            "username": username,
+            "password": password,
+            "proto": proto,
+            "timeout": timeout,
+        }
+        self._session: asynceapi.Device = asynceapi.Device(**session_settings)
```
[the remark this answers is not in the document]
carl-baillargeon: As discussed, keeping type hints for IDE's.

anta/tools.py
```diff
@@ -415,3 +418,45 @@ def format_data(data: dict[str, bool]) -> str:
...
+            else:
+                pending.add(asyncio.create_task(coro))
+
+        if not pending:
+            return
+
+        done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
+
+        while done:
+            yield done.pop()
```
gmuloc, line 462:
```suggestion
    while pending or not coros_ended:
        async for test in coroutines:
            # Add tests to the pending set until the limit is reached or no more tests are available
            if len(pending) < limit:
                # Ensure the coroutine is scheduled to run and add it to the pending set
                pending.add(asyncio.create_task(test))
                logger.debug("Added a test to the pending set: %s", test)
            else:
                logger.debug("Concurrency limit reached: %s tests running. Waiting for tests to complete.", limit)
                break
        if len(pending) < limit and not coros_ended:
            coros_ended = True
            logger.debug("All tests have been added to the pending set.")
        if not pending:
            logger.debug("No pending tests and all tests have been processed. Exiting.")
            return

        done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)

        while done:
            yield done.pop()
```

anta/_runner.py
```diff
@@ -0,0 +1,473 @@
...
+            return None
+        return coroutine
+
+    def _log_run_information(self, *, dry_run: bool = False) -> None:
+        """Log ANTA run information and potential resource limit warnings."""
+        if self._inventory_stats is None:
+            msg = "The inventory stats are not available. ANTA must be executed through AntaRunner.run()"
+            raise RuntimeError(msg)
+
+        width = min(int(console.width) - 34, len("  Total potential connections: 100000000\n"))
```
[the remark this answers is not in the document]
carl-baillargeon: Agreed, let's do it in a separate PR.

anta/tools.py
carl-baillargeon: Done.

docs/faq.md
```diff
@@ -43,11 +43,39 @@ anta_title: Frequently Asked Questions (FAQ)
...
+
+    ANTA also offers several test scheduling strategies to optimize test execution, particularly relevant when the total number of tests exceeds the maximum concurrency limit. It is configurable via the `ANTA_SCHEDULING_STRATEGY` environment variable (default is `round-robin`), along with `ANTA_SCHEDULING_TESTS_PER_DEVICE` (default is **100**) for the `device-by-count` strategy:
+
+    - **Round-robin (`round-robin`)**: Distributes tests evenly across devices. This is generally suitable for small to medium-sized (around 200 devices) fabrics but can open many simultaneous connections.
+    - **Device-by-Device (`device-by-device`)**: Executes all tests for one device before moving on to the next, which may help reduce peak concurrent connections.
+    - **Device-by-Count (`device-by-count`)**: Limits the number of tests scheduled per device in each round. This provides finer control in larger environments where opening too many connections simultaneously might exceed system limits.
```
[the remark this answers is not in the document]
carl-baillargeon: Yes our team is building a lab internally to test this.

tests/units/test_tools.py
carl-baillargeon: Yes I will add/improve unit tests for the runner in a separate PR.

tests/units/test__runner.py
```diff
@@ -0,0 +1,310 @@
...
+        async for coro in anta_runner._test_generator(manager):
+            coro.close()
+
+        # Round-robin between devices
+        assert manager.results[0].name == "leaf1"
+        assert manager.results[1].name == "leaf2"
+        assert manager.results[2].name == "spine1"
+        assert manager.results[3].name == "leaf1"
+        assert manager.results[4].name == "leaf2"
+        assert manager.results[5].name == "spine1"
```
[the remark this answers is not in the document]
carl-baillargeon: Done.

tests/units/test__runner.py
```diff
@@ -0,0 +1,310 @@
...
+            coro.close()
+
+        # Check that indices 0-1 all have name "leaf1"
+        assert all(result.name == "leaf1" for result in manager.results[0:2])
+
+        # Check that indices 2-3 all have name "leaf2"
+        assert all(result.name == "leaf2" for result in manager.results[2:4])
+
+        # Check that indices 4-5 all have name "spine1"
+        assert all(result.name == "spine1" for result in manager.results[4:6])
```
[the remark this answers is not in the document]
carl-baillargeon: Done.

tests/units/conftest.py
```diff
@@ -83,3 +89,64 @@ def yaml_file(request: pytest.FixtureRequest, tmp_path: Path) -> Path:
...
+    - inventory: Inventory file name from the data directory
+    - catalog: Catalog file name from the data directory
+
+    Optional keys:
+    - manager: ResultManager instance
+    - max_concurrency: Maximum concurrency limit
+    - nofile: File descriptor limit
+    """
+    # Import must be inside fixture to prevent circular dependency from breaking CLI tests:
+    # anta.runner -> anta.cli.console -> anta.cli/* (not yet loaded) -> anta.cli.anta
```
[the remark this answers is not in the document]
carl-baillargeon: Yes separate PR.

tests/benchmark/test_runner.py
```diff
@@ -52,3 +59,19 @@ def bench() -> list[Coroutine[Any, Any, TestResult]]:
...
     count = sum(len(tests) for tests in selected_tests.values())
     assert count == len(coroutines)
+
+
+def test_setup_tests(benchmark: BenchmarkFixture, catalog: AntaCatalog, inventory: AntaInventory) -> None:
+    """Benchmark `anta._runner.AntaRunner._setup_tests`."""
+    runner = AntaRunner(inventory=inventory, catalog=catalog)
+    runner._selected_inventory = inventory
+
+    def bench() -> bool:
```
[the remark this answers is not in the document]
carl-baillargeon: Fixed.

tests/units/test__runner.py
```diff
@@ -0,0 +1,310 @@
...
+
+        # The last 3 results should be "leaf1", "leaf2", "spine1" since there is no more tests to run
+        assert manager.results[-3].name == "leaf1"
+        assert manager.results[-2].name == "leaf2"
+        assert manager.results[-1].name == "spine1"
```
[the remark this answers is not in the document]
carl-baillargeon: Fixed.

tests/units/test__runner.py
```diff
@@ -0,0 +1,310 @@
...
+
+        # In dry-run mode, the selected inventory is the original inventory
+        assert anta_runner._selected_inventory is not None
+        assert len(anta_runner._selected_inventory) == len(anta_runner.inventory)
+
+        # In dry-run mode, the inventory stats total should match the original inventory length
+        assert anta_runner._inventory_stats is not None
+        assert anta_runner._inventory_stats.total == len(anta_runner.inventory)
+
+        assert "Dry-run mode, exiting before running the tests." in caplog.records[-1].message
```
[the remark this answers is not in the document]
carl-baillargeon: The manager is tested with `test_multiple_runs_no_manager` and `test_multiple_runs_with_manager`.

docs/faq.md
```diff
@@ -43,11 +43,39 @@ anta_title: Frequently Asked Questions (FAQ)
...
+
+    You can adjust the maximum concurrency limit using the `ANTA_MAX_CONCURRENCY` environment variable. The optimal value depends on your system's CPU usage, memory consumption, and file descriptor limits.
+
+    !!! warning
+
+        Increasing the maximum concurrency limit can lead to system instability if the system is not able to handle the increased load. Monitor system resources and adjust the limit accordingly.
+
+    !!! info "Device Connection Limits"
+
+        Each device is limited to a maximum of **100** concurrent connections. This means that, even if ANTA schedules a high number of tests, each device will only attempt to open up to 100 connections at a time. Furthermore, Arista EOS eAPI is inherently protected against overload and is designed to handle high connection volumes safely.
```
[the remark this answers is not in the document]
carl-baillargeon: Fixed.

docs/faq.md
```diff
@@ -43,11 +43,39 @@ anta_title: Frequently Asked Questions (FAQ)
...
+
+???+ faq "Tests throttling `WARNING` in the logs"
+
+    ANTA is designed to execute many tests concurrently while ensuring system stability. If the total test count exceeds the maximum concurrency limit, tests are throttled to avoid overwhelming the asyncio event loop and exhausting system resources. A `WARNING` message is logged at startup when this occurs.
+
+    By default, ANTA schedules up to **10000** tests concurrently. This default is a balance between performance and stability, but it may not be optimal for every system. If the number of tests exceeds this value, ANTA schedules the first 10000 tests and waits for some tests to complete before scheduling more.
+
+    ### Solution
+
+    You can adjust the maximum concurrency limit using the `ANTA_MAX_CONCURRENCY` environment variable. The optimal value depends on your system's CPU usage, memory consumption, and file descriptor limits.
```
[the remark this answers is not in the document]
carl-baillargeon: Fixed.

anta/cli/nrfu/utils.py
```diff
@@ -50,6 +50,7 @@ def run_tests(ctx: click.Context) -> None:
 
     print_settings(inventory, catalog)
     with anta_progress_bar() as AntaTest.progress:
+        # TODO: Use AntaRunner in ANTA v2.0.0
```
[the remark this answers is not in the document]
carl-baillargeon: Fixed.

anta/device.py
```diff
@@ -329,9 +337,9 @@ def __init__(  # noqa: PLR0913
         name: str | None = None,
         enable_password: str | None = None,
         port: int | None = None,
-        ssh_port: int | None = 22,
+        ssh_port: int = 22,
         tags: set[str] | None = None,
-        timeout: float | None = None,
+        timeout: float | None = 30.0,
```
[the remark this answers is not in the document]
carl-baillargeon: Reverted back.

anta/device.py
```diff
@@ -442,6 +461,14 @@ def _keys(self) -> tuple[Any, ...]:
         """
         return (self._session.host, self._session.port)
 
+    @property
+    def max_connections(self) -> int | None:
+        """Maximum number of concurrent connections allowed by the device. Returns None if not available."""
+        try:
+            return attrgetter("_transport._pool._max_connections")(self._session)
```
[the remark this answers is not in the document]
carl-baillargeon: Updated but we need to ignore mypy and ruff :)

anta/device.py
```diff
@@ -385,15 +392,26 @@ def __init__(  # noqa: PLR0913
...
+            "host": host,
+            "port": port,
+            "username": username,
+            "password": password,
+            "proto": proto,
+            "timeout": timeout,
+        }
+        self._session: asynceapi.Device = asynceapi.Device(**session_settings)
+
+        # Build the SSH connection options
```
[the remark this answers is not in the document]
carl-baillargeon: Old stuff. Reverted back.

anta/device.py
```diff
@@ -385,15 +392,26 @@ def __init__(  # noqa: PLR0913
...
             logger.error(message)
             raise ValueError(message)
+
         self.enable = enable
         self._enable_password = enable_password
-        self._session: asynceapi.Device = asynceapi.Device(host=host, port=port, username=username, password=password, proto=proto, timeout=timeout)
-        ssh_params: dict[str, Any] = {}
+
+        # Build the session settings for the `asynceapi` client
+        session_settings: dict[str, Any] = {
```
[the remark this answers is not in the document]
carl-baillargeon: No, reverted back.

anta/_runner.py
```diff
@@ -0,0 +1,494 @@
...
+    async def _setup_coroutines(self, *, manager: ResultManager | None = None) -> AsyncGenerator[Coroutine[Any, Any, TestResult], None]:
+        """Set up test coroutines from selected tests for the ANTA run.
+
+        It creates an async generator of coroutines which are created by the `test` method of the AntaTest instances.
+        Each coroutine is a test to run.
+
+        Parameters
+        ----------
+        manager
+            An optional ResultManager instance to pre-populate with the test results. Used in `dry_run` mode.
```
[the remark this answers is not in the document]
carl-baillargeon: Fixed.

anta/cli/nrfu/utils.py
```diff
@@ -62,6 +62,9 @@ def run_tests(ctx: click.Context) -> None:
                 dry_run=dry_run,
             )
         )
+        # AntaRunner returns test results as they finish, not necessarily in order
+        # TODO: Add CLI option to choose sorting order
+        ctx.obj["result_manager"].sort(sort_by=["name", "test", "result", "custom_field"])
```
gmuloc, line 67: do we want this in this PR? (other perf impact I would think)

anta/tools.py
gmuloc: after tshooting - the issue in this code is that in the else block we have already consumed the coroutine (test) from the generator but we have not scheduled it as a task so it is never scheduled - hence the warning about a coro not being ever awaited.
the warning just give the name of the coro and by essence python does not duplicate warnings so it is a bit hard to track but the solution is to schedule the task as well in the else block before breaking (and adjust the if / else logic as needed)

anta/cli/nrfu/utils.py
carl-baillargeon: Removed.

gmuloc (commented):
some comments .. not done!

tests/units/test_settings.py
```diff
@@ -0,0 +1,117 @@
...
+        setenvvar.setenv("ANTA_NOFILE", "-1")
+        with pytest.raises(ValidationError):
+            AntaRunnerSettings()
+
+        setenvvar.setenv("ANTA_MAX_CONCURRENCY", "0")
+        with pytest.raises(ValidationError):
+            AntaRunnerSettings()
+
+    @pytest.mark.skipif(os.name == "posix", reason="Run this test on Windows only")
+    def test_file_descriptor_limit_windows(self, caplog: pytest.LogCaptureFixture) -> None:
```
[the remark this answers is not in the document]
carl-baillargeon: There are still some tests using `main()` (which uses AntaRunner under the hood) in `test_runner.py`. I added `# TODO: Move this to AntaRunner tests in ANTA v2.0.0` on top of each.

tests/units/test_runner.py
```diff
@@ -30,25 +30,47 @@
...
+    # On Windows, there is an extra log message when AntaRunner is initialized and
+    # tries to adjust the file descriptor limit.
+    if os.name != "posix":
+        record_tuples = 4
+        record_index = 2
+    else:
+        record_tuples = 3
+        record_index = 1
+
+    assert len(caplog.record_tuples) == record_tuples
```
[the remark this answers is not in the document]
carl-baillargeon: Done in both.

docs/faq.md
```diff
@@ -43,11 +43,35 @@ anta_title: Frequently Asked Questions (FAQ)
...
+
+    You can adjust the maximum concurrency limit using the `ANTA_MAX_CONCURRENCY` environment variable. The optimal value depends on your system CPU usage, memory consumption, and file descriptor limits.
+
+    !!! warning
+
+        Increasing the maximum concurrency limit can lead to system instability if the system is not able to handle the increased load. Monitor system resources and adjust the limit accordingly.
+
+    !!! info "Device Connection Limits"
+
+        Each device is limited to a maximum of **100** concurrent connections. This means that, even if ANTA schedules a high number of tests, it will only attempt to open up to 100 connections at a time towards each device.
```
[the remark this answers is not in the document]
carl-baillargeon: Done.

docs/faq.md
```diff
@@ -43,11 +43,35 @@ anta_title: Frequently Asked Questions (FAQ)
...
+
+    !!! tip
+        If you run ANTA on a large fabric or encounter issues related to resource limits, consider tuning this setting.
```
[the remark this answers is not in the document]
carl-baillargeon: Updated.

anta/settings.py
```diff
@@ -0,0 +1,81 @@
...
+DEFAULT_NOFILE = 16384
+
+
+class AntaRunnerSettings(BaseSettings):
+    """Environment variables for configuring the ANTA runner.
+
+    When initialized, relevant environment variables are loaded. If not set, default values are used.
+
+    On POSIX systems, also adjusts the process's soft limit based on the `ANTA_NOFILE` environment variable
+    while respecting the system's hard limit, meaning the new soft limit cannot exceed the system's hard limit.
```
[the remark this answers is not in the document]
carl-baillargeon: Done.

anta/tools.py
carl-baillargeon: Nice findings. Will use Semaphore instead :)

anta/settings.py
```diff
@@ -0,0 +1,88 @@
...
+DEFAULT_NOFILE = 16384
+"""Default value for the maximum number of open file descriptors for the ANTA process."""
+
+
+class AntaRunnerSettings(BaseSettings):
+    """Environment variables for configuring the ANTA runner.
+
+    When initialized, relevant environment variables are loaded. If not set, default values are used.
+
+    On POSIX systems, also adjusts the process's soft limit based on the `ANTA_NOFILE` environment variable
```
gmuloc, line 32:
```suggestion
    On POSIX systems, also adjusts the process soft limit based on the `ANTA_NOFILE` environment variable
```

anta/settings.py
```diff
@@ -0,0 +1,88 @@
...
+    model_config = SettingsConfigDict(env_prefix="ANTA_")
+
+    nofile: PositiveInt = Field(default=DEFAULT_NOFILE)
+    max_concurrency: PositiveInt = Field(default=DEFAULT_MAX_CONCURRENCY)
+
+    # Computed in post-init
+    _file_descriptor_limit: PositiveInt
+
+    # pylint: disable=arguments-differ
+    def model_post_init(self, context: Any, /) -> None:  # noqa: ANN401
```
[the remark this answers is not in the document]
carl-baillargeon: Fixed.

anta/settings.py
```diff
@@ -0,0 +1,88 @@
...
+        """Post-initialization method to set the file descriptor limit for the current ANTA process."""
+        _ = context
```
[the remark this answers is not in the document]
carl-baillargeon: Fixed.

anta/_runner.py
```diff
@@ -0,0 +1,429 @@
...
+    """
+
+    model_config = ConfigDict(frozen=True, extra="forbid")
+    devices: set[str] | None = None
+    tests: set[str] | None = None
+    tags: set[str] | None = None
+    established_only: bool = True
+
+
+class AntaRunner(BaseModel):
```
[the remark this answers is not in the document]
carl-baillargeon: Removed in latest commits.

anta/_runner.py
```diff
@@ -0,0 +1,429 @@
...
+        Internal settings loaded from environment variables. See the class definition
+        in the `anta.settings` module for details.
+
+    Notes
+    -----
+    After initializing an `AntaRunner` instance, tests should only be executed through
+    the `run()` method. This method manages the complete test lifecycle including setup,
+    execution, and cleanup.
+
+    All internal methods and state (prefixed with `_`) are managed by the `run()` method
```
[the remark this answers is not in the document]
carl-baillargeon: No longer applicable in latest commits.

anta/_runner.py
```diff
@@ -0,0 +1,429 @@
...
+
+    # Run with filters
+    second_run_results = asyncio.run(runner.run(scope=AntaRunnerFilter(tags={"leaf"})))
+    ```
+    """
+
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
+    inventory: AntaInventory
+    catalog: AntaCatalog
+    manager: ResultManager | None = None
```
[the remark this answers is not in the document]
carl-baillargeon: Pydantic class removed.

anta/_runner.py
```diff
@@ -0,0 +1,429 @@
...
+
+        if total_tests == 0:
+            tag_msg = f" matching the tags {filters.tags} " if filters.tags else " "
+            logger.warning(
+                "There are no tests%sto run in the current test catalog and device inventory, please verify your inputs.",
+                tag_msg,
+            )
+            return False
+
+        self._selected_tests = device_to_tests
```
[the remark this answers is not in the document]
carl-baillargeon: Yes. Fixed in latest commits.

anta/_runner.py
```diff
@@ -0,0 +1,429 @@
...
+        self._total_tests = total_tests
+        self._potential_connections = None if not all_have_connections else total_connections
```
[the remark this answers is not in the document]
carl-baillargeon: Yes moved to inventory.

anta/_runner.py
```diff
@@ -0,0 +1,479 @@
...
+    from anta.result_manager.models import TestResult
+
+logger = logging.getLogger(__name__)
+
+
+class AntaRunFilters(BaseModel):
+    """Define filters for an ANTA run.
+
+    Filters determine which devices and tests to include in a run, and how to
+    filter them with tags. This class is used with the `AntaRunner.run()` method.
```
gmuloc, line 41:
```suggestion
    filter them with tags. This class is used by the `AntaRunner.run()` method.
```

anta.json
```diff
@@ -0,0 +1,147 @@
+[
```
gmuloc, line 1: should remove

carl-baillargeon: Done

anta/_runner.py
gmuloc: not relevant anymore for this PR

PR #682, opened 2024-05-16
feat(anta.tests): Optimize VerifyRoutingTableEntry by quering all routes for a vrf.

dlobato:
# Description

Reduce number of commands executed on VerifyRoutingTableEntry by querying all routes of a vrf in one go.

# Checklist:

### DIFF
--- a/anta/tests/routing/generic.py
+++ b/anta/tests/routing/generic.py
@@ -7,7 +7,8 @@
 # mypy: disable-error-code=attr-defined
 from __future__ import annotations
 
-from ipaddress import IPv4Address, ip_interface
+from functools import cache
+from ipaddress import IPv4Address, IPv4Interface
 from typing import ClassVar, Literal
 
 from pydantic import model_validator
@@ -131,7 +132,10 @@ class VerifyRoutingTableEntry(AntaTest):
     name = "VerifyRoutingTableEntry"
     description = "Verifies that the provided routes are present in the routing table of a specified VRF."
     categories: ClassVar[list[str]] = ["routing"]
-    commands: ClassVar[list[AntaCommand | AntaTemplate]] = [AntaTemplate(template="show ip route vrf {vrf} {route}", revision=4)]
+    commands: ClassVar[list[AntaCommand | AntaTemplate]] = [
+        AntaTemplate(template="show ip route vrf {vrf} {route}", revision=4),
+        AntaTemplate(template="show ip route vrf {vrf}", revision=4),
+    ]
 
     class Input(AntaTest.Input):
         """Input model for the VerifyRoutingTableEntry test."""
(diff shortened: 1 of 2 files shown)


carl-baillargeon (commented):
Hey David, thanks for the PR! Did you test this against a large routing table? I remember going back and forth myself when I wrote this test. If I remember correctly I ran into issues with large routing tables and I think it was more efficient sending multiple commands.

dlobato:
Not really, that would be an interesting test. Maybe we can sync and test. Guillaume mentioned this and he proposed to have a parameter collect: [all|unique] that could control the behavior.

carl-baillargeon:
> Not really, that would be an interesting test. Maybe we can sync and test. Guillaume mentioned this and he proposed to have a parameter collect: [all|unique] that could control the behavior.

Tested this with a full routing table and ANTA is in fact crashing. I need to investigate more but when doing the same command with curl or using HTTPX outside of ANTA, it works fine. We can probably reproduce easily with other commands that generate a very large output so that's a separate issue in my opinion.

If we implement a knob to control that behavior, we open the door for users to hit that bug. @gmuloc Thoughts?

carl-baillargeon:
> > Not really, that would be an interesting test. Maybe we can sync and test. Guillaume mentioned this and he proposed to have a parameter collect: [all|unique] that could control the behavior.
> 
> Tested this with a full routing table and ANTA is in fact crashing. I need to investigate more but when doing the same command with curl or using HTTPX outside of ANTA, it works fine. We can probably reproduce easily with other commands that generate a very large output so that's a separate issue in my opinion.
> 
> If we implement a knob to control that behavior, we open the door for users to hit that bug. @gmuloc Thoughts?

Update: I was able to successfuly test the full routing table with ANTA 1.0 but we should still implement a knob in the inputs [ all | unique ] to cover the use case where we need to query a single route against a full routing table.

anta/tests/routing/generic.py
```diff
@@ -142,18 +143,21 @@ class Input(AntaTest.Input):
...
-        """Render the template for each route in the input list."""
-        return [template.render(vrf=self.inputs.vrf, route=route) for route in self.inputs.routes]
+        """Render the template for the input vrf."""
+        return [template.render(vrf=self.inputs.vrf)]
+
+    @staticmethod
+    @cache
+    def ip_interface_ip(route: str) -> IPv4Address:
+        """Return the IP address of the provided ip route with mask."""
+        return IPv4Interface(route).ip
```
carl-baillargeon, line 153: I don't see improvements with a full routing table. Makes sense since every indiviual route is unique, not hitting the cache.

Also, we are spending a lot of time waiting on the actual HTTP response since the output is super big. Let's implement the knob for maximum flexibility.

carl-baillargeon (changes requested):
I don't see improvements with a full routing table. Makes sense since every indiviual route is unique, not hitting the cache.

Also, we are spending a lot of time waiting on the actual HTTP response since the output is super big. Let's implement the knob for maximum flexibility.

carl-baillargeon (approved):

PR #783, opened 2024-08-08
feat(anta): Added strict check to VerifyBGPPeerMPCaps test to verify only mentioned multiprotocol capabilities of a BGP peer should be listed

### LINKED ISSUE #781: Add strict check to VerifyBGPPeerMPCaps test to verify only mentioned multiprotocol capabilities of a BGP peer should be listed.
Test scenario:
Under "multiprotocolCaps" only "ipv4Unicast" should be listed ie. if there are any other capabilities then the test should fail.

vitthalmagadum:
# Description

Added strict check to VerifyBGPPeerMPCaps test to verify only mentioned multiprotocol capabilities of a BGP peer should be listed

Fixes #781 

# Checklist:


anta/tests/routing/bgp.py
```diff
@@ -722,6 +725,8 @@ class BgpPeer(BaseModel):
             """IPv4 address of a BGP peer."""
             vrf: str = "default"
             """Optional VRF for BGP peer. If not provided, it defaults to `default`."""
+            strict: bool = False
+            """Optional check for BGP peer. Verifies the only mentioned capabilities should be listed, otherwise test should fail."""
```
MaheshGSLAB, line 729:
```suggestion
            """Optional check to verifies the only mentioned capabilities should be listed, otherwise test should fail. It defaults to false """
```

anta/tests/routing/bgp.py
```diff
@@ -746,8 +751,17 @@ def test(self) -> None:
                 failures = deep_update(failures, failure)
                 continue
 
-            # Check each capability
+            # Fetching the capabilities output
             bgp_output = get_value(bgp_output, "neighborCapabilities.multiprotocolCaps")
+
+            # Verifies the only mentioned capabilities should be listed in case needed.
```
MaheshGSLAB, line 757: i think we should create some helper functions and if strict is true then verification should done separately(only mentioned  capabilities should present  and it should not look for others. If strict is false then as the exiting code there we can converted into a helper function.

vitthalmagadum: Added helper function for the same.

anta/tests/routing/bgp.py
vitthalmagadum: Updated.

anta/tests/routing/bgp.py
```diff
@@ -144,6 +144,50 @@ def _add_bgp_routes_failure(
...
+    Returns
+    -------
+        dict[str, Any]: A dictionary containing the failure message as per strict check.
+
+    """
+    # Prepare the failure dictionary
+    failure: dict[str, Any] = {}
+
+    # Verifies the only mentioned capabilities should be listed when strict is True.
+    if strict and not set(bgp_output).issubset(set(capabilities)):
```
MaheshGSLAB, line 171: The case where some capabilities are not like advertised plus other than mentioned capabilities are also there is not handled for failure log message.

vitthalmagadum: Handled this check as well.
Thanks!!!

anta/tests/routing/bgp.py
```diff
@@ -722,6 +774,8 @@ class BgpPeer(BaseModel):
             """IPv4 address of a BGP peer."""
             vrf: str = "default"
             """Optional VRF for BGP peer. If not provided, it defaults to `default`."""
+            strict: bool = False
+            """Optional check to verifies the only mentioned capabilities should be listed, otherwise test should fail. It defaults to false."""
```
carl-baillargeon, line 778:
```suggestion
            """If True, requires exact matching of provided capabilities. Defaults to False."""
```

anta/tests/routing/bgp.py
```diff
@@ -685,6 +734,8 @@ def test(self) -> None:
 class VerifyBGPPeerMPCaps(AntaTest):
     """Verifies the multiprotocol capabilities of a BGP peer in a specified VRF.
 
+    Optionally test verifies the only specified multiprotocol capabilities should be listed.
```
carl-baillargeon, line 737:
```suggestion
    Supports `strict: True` to verify that only the specified capabilites are configured, requiring an exact match.
```

anta/tests/routing/bgp.py
```diff
@@ -746,20 +800,14 @@ def test(self) -> None:
...
-                # Check if capabilities are not advertised, received, or enabled
-                elif not all(capability_output.get(prop, False) for prop in ["advertised", "received", "enabled"]):
-                    failure["bgp_peers"][peer][vrf][capability] = capability_output
-                    failures = deep_update(failures, failure)
+
+            # Collecting the failure logs if any.
+            failure_log = _check_peer_capabilities(bgp_output, capabilities, strict=bgp_peer.strict)
+            if failure_log:
+                failure["bgp_peers"][peer][vrf] = failure_log
+                failures = deep_update(failures, failure)
```
carl-baillargeon, line 810: I am not entirely sure if we _should_ use `deep_update` since it's from Pydantic v1. Can you investigate if we can use something else or have a util function doing the same?

anta/tests/routing/bgp.py
```diff
@@ -144,6 +144,55 @@ def _add_bgp_routes_failure(
...
+
+    - Checks the only mentioned capabilities should be listed when strict is True.
+    - Checks the any capability(s) are missing from output.
+    - Checks the any capability(s) are not advertised, received, or enabled
+
+    Parameters
+    ----------
+        bgp_output: The BGP output from the device.
+        capabilities: List of multiprotocol capabilities to be verified.
+        strict: Optional check to verifies the only mentioned capabilities should be listed, otherwise test should fail. It defaults to false
```
carl-baillargeon, line 158:
```suggestion
        strict: If True, requires exact matching of provided capabilities. Defaults to False.
```

anta/tests/routing/bgp.py
```diff
@@ -144,6 +144,55 @@ def _add_bgp_routes_failure(
...
+
+        The * indicates that all arguments following it must be provided as keyword arguments. To compensate boolean-type-hint-positional-argument (FBT001).
+
```
carl-baillargeon, line 161: Not required. We use this everywhere because of Ruff so that's fine.

anta/tests/routing/bgp.py
```diff
@@ -144,6 +144,55 @@ def _add_bgp_routes_failure(
...
+    Returns
+    -------
+        dict[str, Any]: A dictionary containing the failure message as per strict check.
+
+    """
+    # Prepare the failure dictionary
+    failure: dict[str, Any] = {}
+
+    # Verifies the only mentioned capabilities should be listed when strict is True.
+    if strict and not set(bgp_output).issubset(set(capabilities)):
```
carl-baillargeon, line 171: Why are we checking with `issubet`? If strict is True we should have the exact same capabilities no?

anta/tests/routing/bgp.py
```diff
@@ -144,6 +144,55 @@ def _add_bgp_routes_failure(
...
+        other_capabilities = ", ".join(list(set(bgp_output) - set(capabilities)))
+        failure["strict"] = f"Other than mentioned BGP peer multiprotocol capabilities following capability(s) are listed: {other_capabilities}"
```
carl-baillargeon, line 173: What happens if the user provides more capabilities than the actual output? Is that covered here?

anta/tests/routing/bgp.py
```diff
@@ -144,6 +144,55 @@ def _add_bgp_routes_failure(
     return failure_routes
 
 
+def _check_peer_capabilities(bgp_output: dict[str, Any], capabilities: list[MultiProtocolCaps], *, strict: bool = False) -> dict[str, Any]:
```
carl-baillargeon, line 147: We need unit tests for this :)

anta/tests/routing/bgp.py
vitthalmagadum: Addressed

anta/tests/routing/bgp.py
vitthalmagadum: Addressed

anta/tests/routing/bgp.py
vitthalmagadum: Addressed

anta/tests/routing/bgp.py
vitthalmagadum: Updated check with for exact match.

anta/tests/routing/bgp.py
vitthalmagadum: Updated check with for exact match.

anta/tests/routing/bgp.py
vitthalmagadum: Removed the helper function. added check in test.

anta/tests/routing/bgp.py
vitthalmagadum: As It is existing test case, kept this deep update as it is.

carl-baillargeon (approved):
LGTM. Thanks!

PR #805, opened 2024-08-29
feat(anta): Updated VerifyBFDPeersIntervals for failure message to be in milliseconds like the inputs

### LINKED ISSUE #799: VerifyBFDPeersIntervals failure message should be in milliseconds like the inputs
eAPI reports the timers in microseconds instead of milliseconds so we need to convert to be consistent with the inputs.

vitthalmagadum:
# Description

In eAPI reports the interval timers are in microseconds instead of milliseconds so converted timers in to milliseconds to be consistent with the inputs.

Fixes #799 

# Checklist:

### DIFF
--- a/anta/tests/bfd.py
+++ b/anta/tests/bfd.py
@@ -159,8 +159,8 @@ class VerifyBFDPeersIntervals(AntaTest):
             vrf = bfd_peers.vrf
 
             # Converting milliseconds intervals into actual value
-            tx_interval = bfd_peers.tx_interval * 1000
-            rx_interval = bfd_peers.rx_interval * 1000
+            tx_interval = bfd_peers.tx_interval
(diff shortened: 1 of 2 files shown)


anta/tests/bfd.py
```diff
@@ -159,8 +159,8 @@ def test(self) -> None:
             vrf = bfd_peers.vrf
 
             # Converting milliseconds intervals into actual value
```
gmuloc, line 161: this comment would be useless

gmuloc (changes requested):

anta/tests/bfd.py
vitthalmagadum: Updated. Thanks!!

gmuloc (approved):

PR #826, opened 2024-09-12
ci: add codspeed to benchmark ANTA

mtache:
# Description

Use [Codspeed](https://codspeed.io/) to benchmark ANTA.
Use [RESPX](https://lundberg.github.io/respx/) to mock eAPI responses.

The code generates an ANTA catalog from the unit test data at [tests/units/anta_tests](https://github.com/aristanetworks/anta/tree/main/tests/units/anta_tests).
We should consider refactoring these data to use a dictionary with a tuple (Type[AntaTest], [unit test name]: str) as key to avoid duplicates.

It relies on a specific eAPI request ID format: `ANTA-{test name}:{unit test name}:{command index}-{command UUID}`. We currently patch ANTA commands to implement this, but we could consider implementing this directly in ANTA (the eAPI request ID is used for debugging purpose and its format has been arbitrary set to `ANTA-{test name}-{command UUID}`.

# Benchmark performances

As of this writing, there are 556 unit tests for ANTA tests. The code will run 2 benchmarks with 1 device on which those 556 tests are run and another one with 2 devices for a total of 1112 tests run (556 on each device).
This takes 4 minutes to run on the current GitHub Actions runners, `pytest-xdist` (see https://docs.codspeed.io/benchmarks/python#running-benchmarks-in-parallel) has not improved the performances. This could be considered on a bigger runner, but we will loose stdout.

# Requires

- [x] Codspeed application activated on [aristanetworks/anta](https://github.com/aristanetworks/anta)
- [x] #827 

# Checklist:

### DIFF
--- a/anta/custom_types.py
+++ b/anta/custom_types.py
@@ -4,9 +4,9 @@
 """Module that provides predefined types for AntaTest.Input instances."""
 
 import re
-from typing import Annotated, Literal
+from typing import Annotated, Literal, Self, get_args
 
-from pydantic import Field
+from pydantic import BaseModel, Field, model_validator
 from pydantic.functional_validators import AfterValidator, BeforeValidator
 
 # Regular Expression definition
@@ -204,3 +204,36 @@ BgpDropStats = Literal[
 ]
 BgpUpdateError = Literal["inUpdErrWithdraw", "inUpdErrIgnore", "inUpdErrDisableAfiSafi", "disabledAfiSafi", "lastUpdErrTime"]
 BfdProtocol = Literal["bgp", "isis", "lag", "ospf", "ospfv3", "pim", "route-input", "static-bfd", "static-route", "vrrp", "vxlan"]
+
+
+class APISSLCertificate(BaseModel):
+    """Model for an API SSL certificate."""
+
+    certificate_name: str
+    """The name of the certificate to be verified."""
+    expiry_threshold: int
+    """The expiry threshold of the certificate in days."""
+    common_name: str
+    """The common subject name of the certificate."""
+    encryption_algorithm: EncryptionAlgorithm
+    """The encryption algorithm of the certificate."""
+    key_size: RsaKeySize | EcdsaKeySize
+    """The encryption algorithm key size of the certificate."""
+
+    @model_validator(mode="after")
+    def validate_inputs(self) -> Self:
+        """Validate the key size provided to the APISSLCertificates class.
+
+        If encryption_algorithm is RSA then key_size should be in {2048, 3072, 4096}.
+
+        If encryption_algorithm is ECDSA then key_size should be in {256, 384, 521}.
+        """
+        if self.encryption_algorithm == "RSA" and self.key_size not in get_args(RsaKeySize):
+            msg = f"`{self.certificate_name}` key size {self.key_size} is invalid for RSA encryption. Allowed sizes are {get_args(RsaKeySize)}."
+            raise ValueError(msg)
+
+        if self.encryption_algorithm == "ECDSA" and self.key_size not in get_args(EcdsaKeySize):
+            msg = f"`{self.certificate_name}` key size {self.key_size} is invalid for ECDSA encryption. Allowed sizes are {get_args(EcdsaKeySize)}."
+            raise ValueError(msg)
+
+        return self
--- a/anta/custom_types.py
+++ b/anta/custom_types.py
@@ -4,9 +4,9 @@
 """Module that provides predefined types for AntaTest.Input instances."""
 
 import re
-from typing import Annotated, Literal, Self, get_args
+from typing import Annotated, Literal
 
-from pydantic import BaseModel, Field, model_validator
+from pydantic import Field
 from pydantic.functional_validators import AfterValidator, BeforeValidator
 
 # Regular Expression definition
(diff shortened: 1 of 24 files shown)


tests/benchmark/test_anta.py
```diff
@@ -0,0 +1,106 @@
...
+    # Disable logging during ANTA execution to avoid having these function time in benchmarks
+    logging.disable()
+
+    def bench() -> ResultManager:
+        """Need to wrap the ANTA Runner to instantiate a new ResultManger for each benchmark run."""
+        manager = ResultManager()
+        asyncio.run(main(manager, inventory, catalog, dry_run=True))
+        return manager
+
+    manager = benchmark(lambda: bench())  # pylint: disable=unnecessary-lambda
```
carl-baillargeon, line 44: Why can't we just do `manager = benchmark(bench)`?

tests/benchmark/utils.py
```diff
@@ -0,0 +1,170 @@
...
+        """AntaMockEnvironment object will always return a new AntaCatalog object based on the initial parsing.
+
+        This is because AntaCatalog objects store indexes when tests are run and we want a new object each time a test is run.
+        """
+        return copy.deepcopy(self._catalog)
+
+    def _generate_catalog(self) -> tuple[AntaCatalog, dict[tuple[str, str], list[dict[str, Any]]]]:
+        """Generate the `catalog` and `eos_data_catalog` attributes."""
+
+        def import_test_modules() -> Generator[ModuleType, None, None]:
```
carl-baillargeon, line 78: Does this have to be nested here?

tests/benchmark/utils.py
```diff
@@ -0,0 +1,170 @@
...
+        """Mock eAPI response.
+
+        If the eAPI request ID has the format `ANTA-{test name}:{unit test name}:{command index}-{command ID}`,
+        the function will return the eos_data from the unit test case.
+
+        Otherwise, it will mock 'show version' command or raise an Exception.
+        """
+        words_count = 3
+
+        def parse_req_id(req_id: str) -> tuple[str, str, int] | None:
```
carl-baillargeon, line 122: Not sure why but I don't like nested functions :)

tests/benchmark/utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+            eos_data = self.eos_data_catalog[(test_name, unit_test_name)]
+
+            # This could happen if the unit test data is not correctly defined
+            if idx >= len(eos_data):
+                msg = f"Error while generating a mock response for unit test {unit_test_name} of test {test_name}: missing test case in eos_data"
+                raise RuntimeError(msg)
+            result = {"output": eos_data[idx]} if ofmt == "text" else eos_data[idx]
+        elif {"cmd": "show version"} in commands and ofmt == "json":
+            # Mock 'show version' request performed during inventory refresh.
```
carl-baillargeon, line 155: Why aren't we just patching `refresh` does it doesn't send a request?

tests/benchmark/utils.py
mtache: This Generator is only used in this function, so I think yes

tests/benchmark/test_anta.py
```diff
@@ -0,0 +1,106 @@
...
+        """Need to wrap the ANTA Runner to instantiate a new ResultManger for each benchmark run."""
+        manager = ResultManager()
+        asyncio.run(main(manager, inventory, catalog))
+        return manager
+
+    manager = benchmark(lambda: bench())  # pylint: disable=unnecessary-lambda
+    logging.disable(logging.NOTSET)
+
+    if len(catalog.tests) * len(inventory) != len(manager.results):
+        # This could mean duplicates exist.
```
carl-baillargeon, line 82: How can we have duplicates? Would adding a unique ID to ResultOverwrite solve this?

tests/benchmark/utils.py
mtache: Patching `refresh` is applicable to unit testing. Here we want to have a "closer to reality" call stack and measure it. Mocking at eAPI request level is the best we can do.

tests/benchmark/utils.py
mtache: Tastes and colors :)

tests/benchmark/test_anta.py
mtache: Good catch, I got lost when trying different syntaxes

tests/benchmark/test_anta.py
mtache: The code generates an ANTA catalog from the unit test data at tests/units/anta_tests.
The "keys" (more conventional than a real data structure key) is (Type[AntaTest], [unit test name]: str).

When writing this PR, I've caught some copy/paste in those data structures that where duplicated unit tests.
Since we manipulate set() in ANTA code, those duplicated were not ported into the AntaCatalog instance, so I've caught AssertionError that I am letting there for now.

As written in the PR descriptionm, we should consider refactoring these data to use a dictionary with a tuple (Type[AntaTest], : str) as key to avoid duplicates.

Would adding a unique ID to ResultOverwrite solve this? No

tests/units/test_custom_types.py
```diff
@@ -281,3 +285,73 @@ def test_interface_case_sensitivity_uppercase() -> None:
...
+        pytest.param("\\", r"Invalid regex: bad escape \(end of pattern\) at position 0", id="bad escape"),
+    ],
+)
+def test_validate_regex_invalid(str_input: str, error: str) -> None:
+    """Test validate_regex with invalid regex."""
+    with pytest.raises(ValueError, match=error):
+        validate_regex(str_input)
+
+
+class TestAPISSLCertificate:  # pylint: disable=too-few-public-methods
```
carl-baillargeon, line 323: You think it's necessary to test the Pydantic model? We are not doing it for other tests.

anta/custom_types.py
```diff
@@ -204,3 +204,36 @@ def validate_regex(value: str) -> str:
 ]
 BgpUpdateError = Literal["inUpdErrWithdraw", "inUpdErrIgnore", "inUpdErrDisableAfiSafi", "disabledAfiSafi", "lastUpdErrTime"]
 BfdProtocol = Literal["bgp", "isis", "lag", "ospf", "ospfv3", "pim", "route-input", "static-bfd", "static-route", "vrrp", "vxlan"]
+
+
+class APISSLCertificate(BaseModel):
```
carl-baillargeon, line 209: If we start to "extract" the models from the tests themselves (which I agree to make the tests cleaner and lighter), maybe we should put them in a separate module?

tests/units/test_custom_types.py
mtache: I've written this because it was part of a unit test for an AntaTest class and I've moved it here.
I think as soon as we write a custom validator (being a model validator or a field validator) we should test it. Sometimes it is also important for the test logic (like in BGP tests).
But writing tests for every pydantic field is not realistic.

anta/custom_types.py
mtache: In this case, it is used as type in a test. It is just another custom type. I could also have let it in its test and write its unit test in the related unit test for the AntaTest, but this type is quite generic. It is an API SSL Profile for EOS.

anta/custom_types.py
```diff
@@ -4,9 +4,9 @@
 """Module that provides predefined types for AntaTest.Input instances."""
 
 import re
-from typing import Annotated, Literal
+from typing import Annotated, Literal, Self, get_args
```
carl-baillargeon, line 7: `Self` was added in 3.11: https://docs.python.org/3/library/typing.html#typing.Self

tests/units/anta_tests/routing/test_generic.py
```diff
@@ -310,11 +303,14 @@
...
-        "name": "collect-input-error",
-        "test": VerifyRoutingTableEntry,
-        "eos_data": {},
-        "inputs": {"vrf": "default", "routes": ["10.1.0.1", "10.1.0.2"], "collect": "not-valid"},
-        "expected": {"result": "error", "messages": ["Inputs are not valid"]},
-    },
 ]
+
+
+class TestVerifyRoutingTableSize:  # pylint: disable=too-few-public-methods
```
carl-baillargeon, line 309: Same as the other comment, not sure if we should test the Pydantic validation.

anta/custom_types.py
mtache: I am crying

tests/benchmark/utils.py
carl-baillargeon: We have tests that sends `show version`. Would they be impacted by this somehow?

carl-baillargeon (approved):
LGTM

PR #842, opened 2024-09-30
fix(anta.tests): allow for positive or negative offset time zones

Xel-Naha:
# Description

When testing ANTA logging format on switches with timezone Europe/London we got fail on the test VerifyLoggingTimestamp. Looking at the regex it seems to match on negative offsets:

 r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6}-d{2}:\d{2}"

-d{2}:\d{2}"

to allow for positive offsets change this to allow either negative - or positive + 

[\+-]d{2}:\d{2}"


gmuloc:
Hello @Xel-Naha thanks for opening this PR - could you please look at adding some test to make sure we capture this in our CI? If you need pointers let me know!

Xel-Naha:
> Hello @Xel-Naha thanks for opening this PR - could you please look at adding some test to make sure we capture this in our CI? If you need pointers let me know!

hi @gmuloc happy to write a test but some pointers would certainly be helpful,

Xel-Naha (commented):
Added unit test for positive offset and rename original unit test to negative offset

anta/tests/logging.py
```diff
@@ -282,7 +282,7 @@ class VerifyLoggingTimestamp(AntaTest):
     def test(self) -> None:
         """Main test function for VerifyLoggingTimestamp."""
         log_pattern = r"ANTA VerifyLoggingTimestamp validation"
-        timestamp_pattern = r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6}-\d{2}:\d{2}"
+        timestamp_pattern = r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6}[\+-]\d{2}:\d{2}"
```
gmuloc, line 285: quick question - any reason you have added `\` in the list of accepted character to escape `+`?
```suggestion
        timestamp_pattern = r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6}[\+-]\d{2}:\d{2}"
```

Trying out here: https://regex101.com/r/WHhYDT/1 it seems that the result would be the same with and without?

Otherwise LGTM

Xel-Naha: hi @gmuloc I escaped it thinking it was a protected character:

_"+ matches the previous token between one and unlimited times, as many"_

having said that, once it is inside the brackets looks like it treats it like a literal as you say. I'll update and push out again.

Xel-Naha: Pushed the revised code after testing, @gmuloc feel free to review and thanks for showing me, learned something new

gmuloc: thanks both were working but just wanted to make sure if there was a reason or not!

gmuloc (approved):

Xel-Naha:
hi @gmuloc, I just had a look at the failing test and I don't think this is any issue with the code (after all its a minor modification from existing code) however the pipeline seems to have issues downloading something:

```
__getitem__
    raise KeyError(key) from None
KeyError: 'SSS_SERVICE_URL'
Error: Process completed with exit code 1.
Run cat ./target/job_summary.md >> $GITHUB_STEP_SUMMARY
cat: ./target/job_summary.md: No such file or directory
Error: Process completed with exit code 1.
```

i'm suspecting it might be something that needs reviewing on your end?

gmuloc:
> hi @gmuloc, I just had a look at the failing test and I don't think this is any issue with the code (after all its a minor modification from existing code) however the pipeline seems to have issues downloading something:
> 
> ```
> __getitem__
>     raise KeyError(key) from None
> KeyError: 'SSS_SERVICE_URL'
> Error: Process completed with exit code 1.
> Run cat ./target/job_summary.md >> $GITHUB_STEP_SUMMARY
> cat: ./target/job_summary.md: No such file or directory
> Error: Process completed with exit code 1.
> ```
> 
> i'm suspecting it might be something that needs reviewing on your end?

yeah its on our side we need to check with people maintaining this service internally.
we will merge as soon as we can

PR #848, opened 2024-10-02
ci: remove "0x000" calls in codspeed flame graphs

mtache:
# Description

Trying to remove "0x000" calls in codspeed flame graphs. See screenshot below:

Screenshot 2024-10-07 at 16 44 01

They seem to represent an asyncio task.
The name "0x00" changes between benchmarks and prevent comparing code within the asyncio task.
Tried to name the task using https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_task but it did not work.


mtache:
There is not a good solution to that problem today

PR #851, opened 2024-10-04
feat(anta)!: Use HTTP HEAD request instead of a TCP connection to check if device is online

dlobato:
# Description

Use HTTP HEAD request to check connection to a device instead of pure L3 check.
This allows to check connectivity on devices behind a reverse proxy, so rather than checking the reverse proxy is alive using HTTP the request will be forwarded to the right device.

The change is as close to the original as possible, but we could potentially check if the credentials are correct as well.

# Checklist:

### DIFF
--- a/anta/device.py
+++ b/anta/device.py
@@ -539,12 +539,17 @@ class AsyncEOSDevice(AntaDevice):
         """Update attributes of an AsyncEOSDevice instance.
 
         This coroutine must update the following attributes of AsyncEOSDevice:
-        - is_online: When a device IP is reachable and a port can be open
+        - is_online: When a device eAPI HTTP endpoint is accessible
         - established: When a command execution succeeds
         - hw_model: The hardware model of the device
         """
         logger.debug("Refreshing device %s", self.name)
-        self.is_online = await self._session.check_connection()
+        try:
+            self.is_online = await self._session.check_api_endpoint()
+        except HTTPError as e:
+            self.is_online = False
+            logger.warning("Could not connect to device %s: %s", self.name, e)
+
         if self.is_online:
             show_version = AntaCommand(command="show version")
             await self._collect(show_version)
@@ -558,8 +563,6 @@ class AsyncEOSDevice(AntaDevice):
                 # and it is nice to get a meaninfule error message
                 elif self.hw_model == "":
                     logger.critical("Got an empty 'modelName' in the 'show version' returned by device %s", self.name)
-        else:
-            logger.warning("Could not connect to device %s: cannot open eAPI port", self.name)
 
         self.established = bool(self.is_online and self.hw_model)
 
--- a/anta/device.py
+++ b/anta/device.py
@@ -548,23 +548,27 @@ class AsyncEOSDevice(AntaDevice):
             self.is_online = await self._session.check_api_endpoint()
         except HTTPError as e:
             self.is_online = False
+            self.established = False
             logger.warning("Could not connect to device %s: %s", self.name, e)
+            return
 
-        if self.is_online:
-            show_version = AntaCommand(command="show version")
-            await self._collect(show_version)
-            if not show_version.collected:
-                logger.warning("Cannot get hardware information from device %s", self.name)
-            else:
-                self.hw_model = show_version.json_output.get("modelName", None)
-                if self.hw_model is None:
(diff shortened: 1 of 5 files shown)


mtache (changes requested):
Hi David,

_This changes the current behaviour, hence changing the PR name._

I understand that your use case is to actually check that eAPI is reachable behind a reverse proxy which is not possible today with the simple call to `asyncio.open_connection()`.

I checked the behaviour of a HEAD request on the base_url and here is the result:
```
$ curl -k --head https://172.20.1.3
HTTP/1.1 301 Moved Permanently
Server: nginx
Date: Thu, 10 Oct 2024 15:34:22 GMT
Content-Type: text/html
Content-Length: 13
Connection: keep-alive
Location: https://172.20.1.3/eapi/
```

`https://172.20.1.3/eapi/` is actually the eAPI Explorer.
So we can either do that, expect and 301 and test the redirect URL (https://www.python-httpx.org/quickstart/#redirection-and-history)

Or we can do a HEAD on `https://{base_url}/command-api` and expect a 401 if not authenticated:
```
$ curl -k --head https://172.20.1.3/command-api
HTTP/1.1 401 Unauthorized
Server: nginx
Date: Thu, 10 Oct 2024 15:38:59 GMT
Content-Type: text/plain
Content-Length: 30
Connection: keep-alive
```
I assume the code should also expect a 200 if the `asynceapi.Device` instance is authenticated.

Please update the code to handle that behaviour and write unit tests :)

Regarding login, I like the idea of checking the credentials before sending the "show version" eAPI request like we do today. I've opened #871 to track this.

Thanks !

dlobato:
Hi Matthieu,

I also think that requesting headers from command-api would be better. Since we have the credentials we can then check the status code and conclude if we can actually connect to the eAPI server.

I'll update the code and add the required tests.

David

asynceapi/aio_portcheck.py
carl-baillargeon, line 1: Let's keep this module as is to avoid a breaking change.

asynceapi/device.py
```diff
@@ -109,17 +109,20 @@ def __init__(
...
         -------
         bool
-            True when the device eAPI is accessible, False otherwise.
+            True when the device eAPI HTTP endpoint is accessible (2xx status code),
+            otherwise an HTTPStatusError exception is raised.
         """
-        return await port_check_url(self.base_url)
+        response = await self.head(self.EAPI_COMMAND_API_URL, timeout=5)
+        response.raise_for_status()
+        return True
```
carl-baillargeon, line 125: This is a breaking change because we are raising instead or returning False. Let's keep this method as is and create a new one called `check_api_endpoint` with this logic.

anta/device.py
```diff
@@ -513,12 +513,17 @@ async def refresh(self) -> None:
...
         - hw_model: The hardware model of the device
         """
         logger.debug("Refreshing device %s", self.name)
-        self.is_online = await self._session.check_connection()
+        try:
+            self.is_online = await self._session.check_connection()
+        except HTTPError as e:
+            self.is_online = False
+            logger.warning("Could not connect to device %s: %s", self.name, e)
+
```
carl-baillargeon, line 526: Here we can use `self._session.check_api_endpoint()` instead (see my other comment).

carl-baillargeon (commented):
@gmuloc I am ok with merging this (after fixing my review comments) in a minor release but please confirm and consider the following for `refresh`:

1- Currently, `is_online` could be `True` even if subsequent commands failed due to HTTP errors (like 401 Unauthorized). Now, such errors will cause `is_online` to be `False` right away.

2- Errors like 401, 500 are detected _before_ attempting `show version`, whereas before they were detected _after_ the `_collect(show_version)` failed.

3- The warning message associated with these failures changes from potentially `Cannot get hardware information...` (if `_collect` failed) to the new `Could not connect... eAPI endpoint: [Error details]`.

4- Because of this, `self.established` will become `False` earlier and for more reasons (HTTP errors during the initial check) than before.

To me it's arguably an improvement/fix but I will let you chime in.

anta/device.py
dlobato: done

gmuloc (approved):

carl-baillargeon (approved):
LGTM.

PR #858, opened 2024-10-08
refactor(anta): Refactor the VerifyBFDPeersIntervals test to add detection time

### LINKED ISSUE #857: Refactor the VerifyBFDPeersIntervals test to verify the detection time.

vitthalmagadum:
# Description
Verifies the timers of the IPv4 BFD peers in the specified VRF.

    Expected Results
    ----------------
    * Success: The test will pass if the timers of the IPv4 BFD peers are correct in the specified VRF.
    * Failure: The test will fail if the IPv4 BFD peers are not found or their timers are incorrect in the specified VRF.

Fixes #857 

# Checklist:


anta/input_models/bfd.py
```diff
@@ -31,6 +31,9 @@ class BFDPeer(BaseModel):
     """Multiplier of BFD peer. Required field in the `VerifyBFDPeersIntervals` test."""
     protocols: list[BfdProtocol] | None = None
     """List of protocols to be verified. Required field in the `VerifyBFDPeersRegProtocols` test."""
+    detection_time: int | None = None
+    """Detection time of BFD peer in milliseconds. The period of time without receiving BFD packets
+    after which the session is determined to have failed."""
```
carl-baillargeon, line 36:
```suggestion
    """Detection time of BFD peer in milliseconds. Defines how long to wait without receiving BFD packets before declaring the peer session as down.
    
    Optional field in the `VerifyBFDPeersIntervals` test.
    """
```
Please update the test logic + docstring to accomodate this optional field (to avoid breaking changes).

anta/tests/bfd.py
```diff
@@ -151,6 +153,7 @@ def test(self) -> None:
             tx_interval = bfd_peer.tx_interval
             rx_interval = bfd_peer.rx_interval
             multiplier = bfd_peer.multiplier
+            detect_time = bfd_peer.detection_time
```
carl-baillargeon, line 156: We should add a `@field_validator` in Input to check that `tx_interval`, `rx_interval` and `multipler` are provided, similar to what we do with `BgpPeer`. Previous behaviour: https://github.com/aristanetworks/anta/pull/926/files#diff-ea769c2163365318e4e31c317a8c772e6bf95083314abc9ae8c3f493d30b515cL141-L146

`detect_time` should be optional to avoid breaking changes.

carl-baillargeon: We should also update/fix the other BFD tests with proper validators (in a separate PR).

carl-baillargeon (changes requested):

anta/input_models/bfd.py
vitthalmagadum: Updated. Thanks!!

anta/tests/bfd.py
vitthalmagadum: addressed this in `https://github.com/aristanetworks/anta/pull/999`.
Thanks!!

PR #902, opened 2024-10-28
fix(anta.tests): Fix VerifyNTPAssociations to support IP-DNS peers

### LINKED ISSUE #900: Bug: False positive on VerifyNTPAssociation
`show ntp associations | json` can return the following output:
```
DC05-Temp-Bleaf1#sho ntp associations | json
{
    "peers": {
        "10.10.200.22 (ntp.example.com)": {
            "condition": "sys.peer",
            "peerIpAddr": "10.10.200.22",
            "refid": "164.100.255.122",
            "stratumLevel": 3,
            "peerType": "unicast",
            "lastReceived": 1730124132.0,
            "pollInterval": 1024,
            "reachabilityHistory": [
                true,
                true,
                true,
                true,
                true,
                true,
                true,
                true
            ],
            "delay": 0.524,
            "offset": 5.545,
            "jitter": 3.8
        }
    }
}
```
The FQDN of the NTP server can be included in parentheses. The test logic should be updated accordingly.

carl-baillargeon:
# Description

Update VerifyNTPAssociations test to support IP-DNS peers in the JSON output. We now check if the provided NTP server IP or DNS is matching either the peer key or `peerIpAddr` in the details.

Fixes #900 

# Checklist:

### DIFF
--- a/tests/units/anta_tests/test_system.py
+++ b/tests/units/anta_tests/test_system.py
@@ -347,6 +347,39 @@ poll interval unknown
         "expected": {"result": "success"},
     },
     {
+        "name": "success-ip-dns",
+        "test": VerifyNTPAssociations,
+        "eos_data": [
+            {
+                "peers": {
+                    "1.1.1.1 (1.ntp.networks.com)": {
+                        "condition": "sys.peer",
+                        "peerIpAddr": "1.1.1.1",
+                        "stratumLevel": 1,
+                    },
+                    "2.2.2.2 (2.ntp.networks.com)": {
+                        "condition": "candidate",
+                        "peerIpAddr": "2.2.2.2",
+                        "stratumLevel": 2,
+                    },
+                    "3.3.3.3 (3.ntp.networks.com)": {
+                        "condition": "candidate",
+                        "peerIpAddr": "3.3.3.3",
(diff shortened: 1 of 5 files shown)


tests/units/anta_tests/test_system.py
```diff
@@ -346,6 +346,39 @@
         },
         "expected": {"result": "success"},
     },
+    {
+        "name": "success-ip-dns",
+        "test": VerifyNTPAssociations,
+        "eos_data": [
+            {
+                "peers": {
+                    "1.1.1.1 (1.ntp.networks.com)": {
```
MaheshGSLAB, line 355: test will still pass if we have actual address as 1.1.1.11 and input address as 1.1.1.1

MaheshGSLAB: Better if we can check the input address with peerIpAddr

carl-baillargeon: Good catch. Will fix it.

carl-baillargeon: Fixed in c8928bd0085d3e01d5fe77ca99065c22d71cae83

gmuloc (approved):

PR #906, opened 2024-10-30
fix(anta): Implement Pydantic class to replace RawCatalogInput type

titom73:
# Description

Implement a pydantic class to replace typing definition of `RawCatalogInput`. So it can be used as a JSON schema to validate user input like inventory models.

# Checklist:

### DIFF
--- a/anta/catalog.py
+++ b/anta/catalog.py
@@ -22,7 +22,7 @@ from pydantic_core import PydanticCustomError
 from yaml import YAMLError, safe_dump, safe_load
 
 from anta.logger import anta_log_exception
-from anta.models import AntaTest, RawCatalogInputModel
+from anta.models import AntaTest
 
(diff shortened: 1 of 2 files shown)


anta/catalog.py
```diff
@@ -35,8 +35,37 @@
 
 logger = logging.getLogger(__name__)
 
-# { <module_name> : [ { <test_class_name>: <input_as_dict_or_None> }, ... ] }
-RawCatalogInput = dict[str, list[dict[str, Optional[dict[str, Any]]]]]
```
carl-baillargeon, line 39: Removing this will break AVD. We use that type here: https://github.com/aristanetworks/avd/blob/devel/ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py#L170

PR #908, opened 2024-10-31
chore: Add vscode python debugger for anta code

titom73:
# Description

Implement VSCode Python debugger configuration to run ANTA CLI in debug mode.

This mechanism gives you following options:

- Add breakpoint to stop code execution
- Variables watch to inspect content during code execution.
- Capture Exceptions

And [more](https://code.visualstudio.com/docs/python/debugging)

Default configuration is based on the following files:

- `.vscode/inventory.yml`: List of hosts to use in anta. Use ATD IP address
  schema.
- `.vscode/nrfu.yml`: List of NRFU tests to run.

# Checklist:


.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
...
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Python Debugger: Module",
+            "type": "debugpy",
+            "console": "integratedTerminal",
+            "request": "launch",
+            "module": "anta.cli._main",
```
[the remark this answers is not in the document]
titom73: There is 2 things here:

- Module is not built to be executed on its own. So to use `anta.cli` only, we need to rename `_main.py` to `__main__.py`

- To execute module, we need to get a default `if __name__ == "__main__"` stanza to load the correct entrypoint.

The thing is VScode debugger uses module approach to monitor code execution, so if we want to get access to it, we need to format our CLI module to be compliant with Python standard structure.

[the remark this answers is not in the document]
gmuloc: it is there: https://github.com/aristanetworks/anta/blob/main/anta/cli/__init__.py#L40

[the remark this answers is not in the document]
gmuloc:
```
ANTA_DEBUG=True python anta/cli/__init__.py
The ANTA command line client could not run because the required dependencies were not installed.
Make sure you've installed everything with: pip install 'anta[cli]'
The caught exception was: attempted relative import with no known parent package
```

ok it is unhappy lets fix our code rather than adding extra entrypoints where we bypass validation

suggested fix in  https://github.com/aristanetworks/anta/pull/909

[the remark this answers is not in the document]
gmuloc: > Module is not built to be executed on its own. So to use anta.cli only, we need to rename _main.py to __main__.py
we don't want this as it bypasses the check for cli dependencies being installed which is not great.


> To execute module, we need to get a default if __name__ == "__main__" stanza to load the correct entrypoint.
we have this in __init__.py but it makes python not happy ... so lets fix it!

.vscode/inventory.yml
```diff
@@ -0,0 +1,21 @@
+---
+anta_inventory:
```
gmuloc, line 2: do we want these in the repo or should it be to each its own?

.vscode/settings.json
```diff
@@ -1,14 +0,0 @@
-{
-  "ruff.enable": true,
```
gmuloc, line 2: Hmm @mtache ok for you if this goes?

tests/units/cli/test__init__.py
```diff
@@ -34,10 +35,11 @@ def test_cli_error_missing(capsys: pytest.CaptureFixture[Any]) -> None:
     """Test ANTA errors out when anta[cli] was not installed."""
     with patch.dict(sys.modules) as sys_modules, patch("builtins.__import__", import_mock):
         del sys_modules["anta.cli._main"]
+        del sys_modules["anta.cli.console"]
```
gmuloc, line 38: is this one needed?

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
...
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Python Debugger: Module",
+            "type": "debugpy",
+            "console": "integratedTerminal",
+            "request": "launch",
+            "module": "anta.cli",
+            "args": [ "nrfu", "-c", "${workspaceFolder}/.vscode/nrfu.yml", "-i", "${workspaceFolder}/.vscode/inventory.yml", "text"],
```
carl-baillargeon, line 13:
```suggestion
            "args": [ "nrfu", "text"],
```
I think we should let the user run NRFU with his own environment variables. We wouldn't need `env` below and maintain the `inventory.yml` + `nrfu.yml` files.

anta/cli/__main__.py
```diff
@@ -0,0 +1,44 @@
...
+
+            def wrap() -> None:
+                """Error message if any CLI dependency is missing."""
+                logging.error(
+                    "The ANTA command line client could not run because the required "
+                    "dependencies were not installed.\nMake sure you've installed "
+                    "everything with: pip install 'anta[cli]'"
+                )
+                if __DEBUG__:
+                    logging.debug("The caught exception was: %s", exception)
```
carl-baillargeon, line 32: This is no longer being triggered as before:
BEFORE:
```
(.pyavd_cli) ~/git_projects/anta (main ✗) anta nrfu json 
The ANTA command line client could not run because the required dependencies were not installed.
Make sure you've installed everything with: pip install 'anta[cli]'
(.pyavd_cli) ~/git_projects/anta (main ✗) ANTA_DEBUG=true anta nrfu json 
The ANTA command line client could not run because the required dependencies were not installed.
Make sure you've installed everything with: pip install 'anta[cli]'
The caught exception was: cannot import name 'ResultManager' from partially initialized module 'anta.result_manager.models' (most likely due to a circular import) (/home/cbaillar/git_projects/anta/anta/result_manager/models.py)
```
AFTER:
```
(.pyavd_cli) ~/git_projects/anta (chore/python-debugger ✗) ANTA_DEBUG=true anta nrfu json
  File "/home/cbaillar/.pyenv/versions/3.12.3/envs/.anta/bin/anta", line 5, in <module>
    from anta.cli import cli
ImportError: cannot import name 'cli' from 'anta.cli' (/home/cbaillar/git_projects/anta/anta/cli/__init__.py)
```

Before we were going `from ._main import anta, cli`, and now `from ._main import cli` - cli is imported first so maybe that's causing the different behavior.

carl-baillargeon: BUT, when running with this new debugger, we get a pretty nice output:
image

.vscode/inventory.yml
titom73: Fixed by removing files and introducing management via `envFile`

.vscode/settings.json
titom73: Reintroduce file

tests/units/cli/test__init__.py
titom73: Work in progress to fix pytest error

.vscode/launch.json
titom73: Fixed by introducing management via `envFile`

PR #913, opened 2024-11-04
refactor: Deprecate the ability to configure aaa config when collecting show-techs

### LINKED ISSUE #894: Deprecate AAA configuration show collect show-tech

gmuloc:
# Description

ANTA should not configure the devices - only outputing a nice error message when something is missing

Fixes #894 

# Checklist:

### DIFF
--- a/anta/cli/exec/utils.py
+++ b/anta/cli/exec/utils.py
@@ -129,7 +129,7 @@ async def collect_show_tech(inv: AntaInventory, root_dir: Path, *, configure: bo
                     return
 
                 # TODO: ANTA 2.0.0
-                msg = "[DEPRECATED] Using '--configure' for collecting show-techs is deprecated and will be removed in ANTA 2.0.0"
+                msg = "[DEPRECATED] Using '--configure' for collecting show-techs is deprecated and will be removed in ANTA 2.0.0. Please add the required configuration on your devices before running this command. from ANTA."
                 logger.warning(msg)
 
                 commands = []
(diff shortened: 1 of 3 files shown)


anta/cli/exec/utils.py
```diff
@@ -128,6 +128,10 @@ async def collect(device: AntaDevice) -> None:
                     logger.error("Unable to collect tech-support on %s: configuration 'aaa authorization exec default local' is not present", device.name)
                     return
 
+                # TODO: ANTA 2.0.0
+                msg = "[DEPRECATED] Using '--configure' for collecting show-techs is deprecated and will be removed in ANTA 2.0.0"
```
carl-baillargeon, line 132: Should we propose the alternative here (configure it yourself)?

anta/cli/exec/utils.py
```diff
@@ -128,6 +128,10 @@ async def collect(device: AntaDevice) -> None:
...
+                logger.warning(msg)
```
carl-baillargeon, line 133: I think we should use `warnings.warn()` here so that users can control these warnings with `warnings.filterwarnings()`

gmuloc: no because by default warnings.warn deprecation warnings are disabled

gmuloc: https://docs.python.org/3/library/warnings.html#default-warning-filter

anta/cli/exec/utils.py
gmuloc, line 132:
```suggestion
                msg = "[DEPRECATED] Using '--configure' for collecting show-techs is deprecated and will be removed in ANTA 2.0.0. Please add the required configuration on your devices before running this command. from ANTA."
```

carl-baillargeon (approved):
LGTM.

PR #937, opened 2024-11-26
feat(anta): add anta.result_manager.models.AtomicTestResult

mtache:
# Description

Fixes # (issue id)

# Checklist:

### DIFF
--- a/anta/models.py
+++ b/anta/models.py
@@ -426,13 +426,14 @@ class AntaTest(ABC):
         self.device: AntaDevice = device
         self.inputs: AntaTest.Input
         self.instance_commands: list[AntaCommand] = []
+        self._init_inputs(inputs)
         self.result: TestResult = TestResult(
             name=device.name,
             test=self.name,
+            inputs=self.inputs,
             categories=self.categories,
             description=self.description,
         )
-        self._init_inputs(inputs)
         if self.result.result == AntaTestStatus.UNSET:
             self._init_commands(eos_data)
 
--- a/anta/result_manager/models.py
+++ b/anta/result_manager/models.py
@@ -5,10 +5,11 @@
 
 from __future__ import annotations
 
+from abc import ABC
 from dataclasses import dataclass, field
 from enum import Enum
 
-from pydantic import BaseModel
+from pydantic import BaseModel, SerializeAsAny
 
 
 class AntaTestStatus(str, Enum):
@@ -28,35 +29,19 @@ class AntaTestStatus(str, Enum):
         return self.value
 
 
-class TestResult(BaseModel):
-    """Describe the result of a test from a single device.
+class BaseTestResult(BaseModel, ABC):
+    """Base model for test results.
 
     Attributes
     ----------
-    name : str
-        Name of the device where the test was run.
-    test : str
-        Name of the test run on the device.
-    categories : list[str]
-        List of categories the TestResult belongs to. Defaults to the AntaTest categories.
-    description : str
-        Description of the TestResult. Defaults to the AntaTest description.
     result : AntaTestStatus
-        Result of the test. Must be one of the AntaTestStatus Enum values: unset, success, failure, error or skipped.
+        Result of the test.
     messages : list[str]
-        Messages to report after the test, if any.
-    custom_field : str | None
-        Custom field to store a string for flexibility in integrating with ANTA.
-
+        Messages reported by the test.
     """
 
-    name: str
-    test: str
-    categories: list[str]
-    description: str
     result: AntaTestStatus = AntaTestStatus.UNSET
     messages: list[str] = []
-    custom_field: str | None = None
 
     def is_success(self, message: str | None = None) -> None:
         """Set status to success.
@@ -111,15 +96,82 @@ class TestResult(BaseModel):
             Status of the test.
         message
             Optional message.
-
         """
         self.result = status
         if message is not None:
             self.messages.append(message)
 
+
+class AtomicTestResult(BaseTestResult):
+    """Describe the result of an atomic test part of a larger test related to a TestResult instance.
+
+    Attributes
+    ----------
+    description : str | None
+        Description of the AtomicTestResult.
+    inputs: BaseModel | None
+        If this AtomicTestResult is related to a specific parent test input, this field must be set.
+    result : AntaTestStatus
+        Result of the atomic test.
+    messages : list[str]
+        Messages reported by the test.
+    """
+
+    description: str | None = None
+    inputs: BaseModel | None = None
+
+
+class TestResult(BaseTestResult):
+    """Describe the result of a test from a single device.
+
+    Attributes
+    ----------
+    name : str
+        Name of the device on which the test was run.
+    test : str
+        Name of the AntaTest subclass.
+    inputs: BaseModel
+        Inputs of the AntaTest instance.
+    categories : list[str]
+        List of categories the TestResult belongs to. Defaults to the AntaTest subclass categories.
+    description : str
+        Description of the TestResult. Defaults to the AntaTest subclass description.
+    result : AntaTestStatus
+        Result of the test.
+    messages : list[str]
+        Messages reported by the test.
+    custom_field : str | None
+        Custom field to store a string for flexibility in integrating with ANTA.
+    atomic_results: list[AtomicTestResult]
+        A list of AtomicTestResult instances which can be used to store atomic results during the test execution.
+        It can then be leveraged in the report to render atomic results over the test global TestResult.
+    """
+
+    name: str
+    test: str
+    description: str
+    inputs: SerializeAsAny[BaseModel]
+    categories: list[str]
+    custom_field: str | None = None
+    atomic_results: list[AtomicTestResult] = []
+
     def __str__(self) -> str:
         """Return a human readable string of this TestResult."""
         return f"Test '{self.test}' (on '{self.name}'): Result '{self.result}'\nMessages: {self.messages}"
+
+    def add(self, description: str | None = None, inputs: BaseModel | None = None) -> AtomicTestResult:
+        """Create and add a new AtomicTestResult to this TestResult instance.
+
+        Parameters
+        ----------
+        description : str | None
+            Description of the AtomicTestResult.
+        inputs: BaseModel | None
+            If this AtomicTestResult is related to a specific parent test input, this field must be set.
+        """
+        res = AtomicTestResult(description=description, inputs=inputs)
+        self.atomic_results.append(res)
+        return res
 
 
 # Pylint does not treat dataclasses differently: https://github.com/pylint-dev/pylint/issues/9058
--- a/tests/units/anta_tests/__init__.py
+++ b/tests/units/anta_tests/__init__.py
@@ -24,10 +24,22 @@ def test(device: AntaDevice, data: dict[str, Any]) -> None:
     assert test_instance.result.result == data["expected"]["result"], f"Expected '{data['expected']['result']}' result, got '{test_instance.result.result}'"
     if "messages" in data["expected"]:
         # We expect messages in test result
-        assert len(test_instance.result.messages) == len(data["expected"]["messages"])
+        assert len(test_instance.result.messages) == len(
+            data["expected"]["messages"]
+        ), f"Expected {len(data["expected"]["messages"])} messages, got {len(test_instance.result.messages)}"
         # Test will pass if the expected message is included in the test result message
         for message, expected in zip(test_instance.result.messages, data["expected"]["messages"]):  # NOTE: zip(strict=True) has been added in Python 3.10
             assert expected in message
     else:
         # Test result should not have messages
-        assert test_instance.result.messages == []
+        assert test_instance.result.messages == [], "There are untested messages"
+
+    if "atomic_results" in data["expected"]:
+        assert len(test_instance.result.atomic_results) == len(
+            data["expected"]["atomic_results"]
+        ), f"Expected {len(data["expected"]["atomic_results"])} atomic results, got {len(test_instance.result.atomic_results)}"
+        for result, expected in zip(test_instance.result.atomic_results, data["expected"]["atomic_results"]):
+            assert result.model_dump(serialize_as_any=True, mode="json") == expected
+    else:
+        # Test result should not have atomic results
+        assert test_instance.result.atomic_results == [], "There are untested atomic results"


anta/cli/nrfu/__init__.py
```diff
@@ -57,6 +57,15 @@ def parse_args(self, ctx: click.Context, args: list[str]) -> list[str]:
...
 @inventory_options
 @catalog_options
+@click.option(
+    "--expand-atomic",
+    "-x",
+    default=False,
+    show_envvar=True,
+    is_flag=True,
+    show_default=True,
+    help="Flag to indicate if atomic results should be rendered",
```
carl-baillargeon, line 67: Will we support all `anta nrfu` commands (csv, json, etc.)? We should be more specific here.

anta/cli/nrfu/utils.py
```diff
@@ -90,8 +90,10 @@ def print_table(ctx: click.Context, group_by: Literal["device", "test"] | None =
         console.print(reporter.report_summary_devices(results))
     elif group_by == "test":
         console.print(reporter.report_summary_tests(results))
+    elif ctx.obj["expand_atomic"]:
+        console.print(reporter.report_expanded(results))
     else:
-        console.print(reporter.report_all(results))
+        console.print(reporter.report(results))
```
carl-baillargeon, line 96: Add this logic to other functions as well if we want to support it.

anta/result_manager/__init__.py
```diff
@@ -137,6 +137,11 @@ def results_by_status(self) -> dict[AntaTestStatus, list[TestResult]]:
         """A cached property that returns the results grouped by status."""
         return {status: [result for result in self._result_entries if result.result == status] for status in AntaTestStatus}
 
+    @cached_property
+    def results_by_category(self) -> list[TestResult]:
```
carl-baillargeon, line 141: Will need to clear it in the `add()` method.

anta/result_manager/models.py
```diff
@@ -111,16 +104,103 @@ def _set_status(self, status: AntaTestStatus, message: str | None = None) -> Non
...
+        Description of the TestResult. Defaults to the AntaTest subclass description.
+    result : AntaTestStatus
+        Result of the test.
+    messages : list[str]
+        Messages reported by the test.
+    custom_field : str | None
+        Custom field to store a string for flexibility in integrating with ANTA.
+    atomic_results: list[AtomicTestResult]
+        A list of AtomicTestResult instances which can be used to store atomic results during the test execution.
+        It can then be leveraged in the report to render atomic results over the test global TestResult.
```
carl-baillargeon, line 177: When serializing a TestResult, the order of the fields seem to be following the class hierchary, `description`, `result` are at the top which is probably not what we want.

anta/reporter/__init__.py
```diff
@@ -102,35 +103,84 @@ def _color_result(self, status: AntaTestStatus) -> str:
...
+            categories = device = test = None
+            if isinstance(result, TestResult):
+                categories = ", ".join(convert_categories(result.categories))
+                device = str(result.name)
+                test = result.test
+            else:
+                test = name
+            state = self._color_result(result.result)
+            message = self._split_list_to_txt_list(result.messages) if len(result.messages) > 0 else ""
+            inputs = result.inputs.model_dump_json(indent=2, exclude={"result_overwrite", "filters"}) if result.inputs is not None else None
```
carl-baillargeon, line 137: I feel like like having all inputs at the parent test level is unnesserary when we have atomic results. We are just repeating them. Maybe we should add them only if atomic results are missing.

anta/cli/nrfu/__init__.py
mtache: I moved this under `anta nrfu` for now.
We support Table, JSON and Jinja for now. Only Table required this option.

anta/cli/nrfu/utils.py
mtache: I think it just makes sense to provide an option for Table and Text.
Other reports needs the atomic results anyway.

anta/reporter/__init__.py
mtache: Actually it depends on the test implementation. One can decide to have atomic results but no inputs tied to these atomic results.
Also, the parent test can have more input than the union of all the atomic results inputs.
But in most test implementation that will be redundant. Let's just remove the parent test inputs.

anta/result_manager/__init__.py
mtache: Thanks!

anta/result_manager/models.py
mtache: There is no way to do that with field inheritance.
So I guess we will redefine the fields in both models.

anta/result_manager/models.py
```diff
@@ -102,6 +81,34 @@ def is_error(self, message: str | None = None) -> None:
...
+        If this AtomicTestResult is related to a specific parent test input, this field must be set.
+    result : AntaTestStatus
+        Result of the atomic test.
+    messages : list[str]
+        Messages reported by the test.
+    """
+
+    _parent: TestResult
+    description: str
+    inputs: SerializeAsAny[InstanceOf[BaseModel]] | None = None
```
gmuloc: could it not be `AntaTest.Inputs`?

anta/result_manager/models.py
```diff
@@ -102,6 +81,34 @@ def is_error(self, message: str | None = None) -> None:
...
+    result: AntaTestStatus = AntaTestStatus.UNSET
+    messages: list[str] = []
+
+    def __init__(self, **data: Any) -> None:  # noqa: ANN401
+        """Instantiate the parent TestResult private attribute."""
+        parent = data.pop("parent")
```
gmuloc, line 108: this will raise a. KeyError if no `parent` is in the data

anta/result_manager/models.py
```diff
@@ -111,16 +118,83 @@ def _set_status(self, status: AntaTestStatus, message: str | None = None) -> Non
             Status of the test.
         message
             Optional message.
-
         """
         self.result = status
+        if (self._parent.result == AntaTestStatus.UNSET and status == AntaTestStatus.SUCCESS) or status in [AntaTestStatus.FAILURE, AntaTestStatus.ERROR]:
```
gmuloc, line 123: lets document this behavior in the docstring

anta/result_manager/models.py
```diff
@@ -111,16 +118,83 @@ def _set_status(self, status: AntaTestStatus, message: str | None = None) -> Non
...
+    description: str
+    inputs: SerializeAsAny[InstanceOf[BaseModel]] | None = None  # A TestResult inputs can be None in case of inputs validation error
+    custom_field: str | None = None
+    result: AntaTestStatus = AntaTestStatus.UNSET
+    messages: list[str] = []
+    atomic_results: list[AtomicTestResult] = []
 
     def __str__(self) -> str:
         """Return a human readable string of this TestResult."""
         return f"Test '{self.test}' (on '{self.name}'): Result '{self.result}'\nMessages: {self.messages}"
```
gmuloc, line 168: shoyld this change to indicate it has AtomicTestResults?

anta/tests/connectivity.py
```diff
@@ -66,11 +66,11 @@ def render(self, template: AntaTemplate) -> list[AntaCommand]:
     @AntaTest.anta_test
     def test(self) -> None:
         """Main test function for VerifyReachability."""
-        self.result.is_success()
-
         for command, host in zip(self.instance_commands, self.inputs.hosts):
+            host_result = self.result.add(description=str(host), inputs=host)
```
gmuloc, line 70: should we find a way to signal a test returns atomic test result for documentation puproses?

docs/contribution.md
```diff
@@ -94,43 +94,63 @@ All submodule should have its own pytest section under `tests/units/anta_tests/<
...
+    result: Literal["success", "failure", "skipped"]  # The expected status of this atomic result.
+    description: str  # The expected description of this atomic result.
+    messages: NotRequired[list[str]]  # The expected messages of this atomic result. The strings can be a substrings of the actual messages.
+    inputs: NotRequired[dict[str, Any]]  # The inputs registered with this atomic result.
+
+
+class Expected(TypedDict):
+    """Expected result of a unit test of an AntaTest subclass."""
+
+    result: Literal["success", "failure", "skipped"]  # The expected status of this unit test.
```
gmuloc, line 122: should it not have more (unset / error)? or is it used only in places where we expect these 3?

docs/contribution.md
```diff
@@ -143,12 +163,94 @@ DATA: list[dict[str, Any]] = [
...
+        "expected": {
+            "result": "failure",
+            "messages": ["Unreachable Destination 10.0.0.11 from 10.0.0.5 in VRF default"],
+            # This test has implemented atomic results.
+            # Expected atomic results must be specified or the test will fail.
+            "atomic_results": [
+                {
+                    # Expected atomic result description
+                    "description": "Destination 10.0.0.11 from 10.0.0.5 in VRF default",
+                    # If the atomic result is tied to a subset of the test inputs, it needs to be expected here.
```
gmuloc, line 221:
```suggestion
                    # If the atomic result is tied to a subset of the test inputs, it needs to be added here otherwise the test will fail.
```

docs/contribution.md
```diff
@@ -143,12 +163,94 @@ DATA: list[dict[str, Any]] = [
...
+
+                """,
+                ],
+            },
+        ],
+        "expected": {
+            "result": "failure",
+            "messages": ["Unreachable Destination 10.0.0.11 from 10.0.0.5 in VRF default"],
+            # This test has implemented atomic results.
+            # Expected atomic results must be specified or the test will fail.
```
gmuloc, line 216: order matters I suppose?

docs/contribution.md
```diff
@@ -143,12 +163,94 @@ DATA: list[dict[str, Any]] = [
...
+                        "destination": "10.0.0.11",
+                        "df_bit": False,
+                        "repeat": 2,
+                        "size": 100,
+                        "source": "10.0.0.5",
+                        "vrf": "default",
+                    },
+                    # Expected atomic result status
+                    "result": "failure",
+                    # If the atomic result returns messages, it needs to be expected otherwise test will fail.
```
gmuloc, line 232:
```suggestion
                    # If the atomic result returns messages, it needs to be added here otherwise test will fail.
```

pyproject.toml
```diff
@@ -417,7 +418,7 @@ runtime-evaluated-base-classes = ["pydantic.BaseModel", "anta.models.AntaTest.In
 ]
 "anta/cli/*" = [
   "PLR0913",  # CLI has many arguments defined in functions
-  "ANN401",   # TODO: Check if we can update the Any type hints in the CLI
+  "FBT001",   # CLI can have boolean as argument
```
gmuloc, line 421: hmm can we not make the argument a keyword-only argument instead?

https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/

tests/benchmark/test_runner.py
```diff
@@ -41,9 +41,10 @@ def test_get_coroutines(benchmark: BenchmarkFixture, catalog: AntaCatalog, inven
     selected_tests = prepare_tests(inventory=inventory, catalog=catalog, tests=None, tags=None)
 
     assert selected_tests is not None
+    results = ResultManager()
```
gmuloc, line 44: is this not making the ResultManager unique across all runs?

tests/benchmark/utils.py
```diff
@@ -90,7 +90,7 @@ def import_test_modules() -> Generator[ModuleType, None, None]:
             for test_data in module.DATA:
                 test = test_data["test"]
                 result_overwrite = AntaTest.Input.ResultOverwrite(custom_field=test_data["name"])
-                if test_data["inputs"] is None:
+                if "inputs" not in test_data or test_data["inputs"] is None:
```
gmuloc, line 93:
```suggestion
                if test_data.get("inputs") is None:
```

tests/units/anta_tests/__init__.py
```diff
@@ -3,10 +3,43 @@
...
+    from typing_extensions import NotRequired  # NOTE: required to support Python < 3.11 https://peps.python.org/pep-0655/#usage-in-python-3-11
+
+    from anta.device import AntaDevice
+    from anta.models import AntaTest
+
+
+class AtomicResult(TypedDict):
+    """Expected atomic result of a unit test of an AntaTest subclass."""
+
+    result: Literal["success", "failure", "skipped"]  # TODO: Refactor tests and use AntaTestStatus
```
gmuloc, line 21: this can be given to one of the GSLabs engineer if clearly defined in an issue

tests/units/anta_tests/__init__.py
```diff
@@ -3,10 +3,43 @@
...
+    messages: NotRequired[list[str]]
+    atomic_results: NotRequired[list[AtomicResult]]
+
+
+class AntaUnitTest(TypedDict):
+    """The parameters required for a unit test of an AntaTest subclass."""
+
+    name: str  # TODO: Refactor tests and change the DATA constant type as dictionary instead of list[AntaUnitTest] to avoid test duplicates.
```
gmuloc, line 38: same - do we have lots of duplicates?

tests/units/anta_tests/__init__.py
```diff
@@ -17,17 +50,37 @@ def test(device: AntaDevice, data: dict[str, Any]) -> None:
     See `tests/units/anta_tests/README.md` for more information on how to use it.
     """
     # Instantiate the AntaTest subclass
+    if "inputs" not in data:
+        data["inputs"] = None
     test_instance = data["test"](device, inputs=data["inputs"], eos_data=data["eos_data"])
```
gmuloc, line 55:
```suggestion
    test_instance = data["test"](device, inputs=data.get("inputs"), eos_data=data["eos_data"])
```

tests/units/anta_tests/__init__.py
```diff
@@ -17,17 +50,37 @@ def test(device: AntaDevice, data: dict[str, Any]) -> None:
...
+        assert len(test_instance.result.messages) == len(
+            data["expected"]["messages"]
+        ), f"Expected {len(data['expected']['messages'])} messages, got {len(test_instance.result.messages)}"
         # Test will pass if the expected message is included in the test result message
         for message, expected in zip(test_instance.result.messages, data["expected"]["messages"]):  # NOTE: zip(strict=True) has been added in Python 3.10
             assert expected in message
     else:
         # Test result should not have messages
-        assert test_instance.result.messages == []
+        assert test_instance.result.messages == [], "There are untested messages"
```
gmuloc, line 70: not necessarily true - there could be messages because the test failed when it should have succeeded. 

```suggestion
        assert test_instance.result.messages == [], "There are unexpected messages"
```

tests/units/reporter/test__init__.py
```diff
@@ -84,65 +84,67 @@ def test__color_result(self, status: AntaTestStatus, expected_status: str) -> No
         assert report._color_result(status) == expected_status
 
     @pytest.mark.parametrize(
-        ("title", "number_of_tests", "expected_length"),
+        ("results_size"),
         [
-            pytest.param(None, 5, 5, id="all results"),
-            pytest.param(None, 0, 0, id="result for host1 when no host1 test"),
-            pytest.param(None, 5, 5, id="result for test VerifyTest3"),
-            pytest.param("Custom title", 5, 5, id="Change table title"),
```
gmuloc, line 92: will probably need to add this back to be non breaking

tests/units/result_manager/conftest.py
```diff
@@ -56,30 +57,35 @@ def result_manager() -> ResultManager:
 
 
 @pytest.fixture
-def test_result_factory(device: AntaDevice) -> Callable[[int], TestResult]:
+def test_result_factory(device: AntaDevice) -> Callable[[int, int, bool, bool], TestResult]:
     """Return a anta.result_manager.models.TestResult object."""
     # pylint: disable=redefined-outer-name
 
-    def _create(index: int = 0) -> TestResult:
+    def _create(index: int = 0, nb_atomic_results: int = 0, distinct_tests: bool = False, distinct_devices: bool = False) -> TestResult:
```
gmuloc, line 64: can we document the new flags in docstring?

tests/units/result_manager/conftest.py
```diff
@@ -56,30 +57,35 @@ def result_manager() -> ResultManager:
...
+        for i in range(nb_atomic_results):
+            res.add(description=f"{test.name}{index}AtomicTestResult{i}", inputs=test.inputs)
+        return res
 
     return _create
 
 
 @pytest.fixture
-def list_result_factory(test_result_factory: Callable[[int], TestResult]) -> Callable[[int], list[TestResult]]:
+def list_result_factory(test_result_factory: Callable[[int, int, bool, bool], TestResult]) -> Callable[[int, int, bool, bool], list[TestResult]]:
```
gmuloc, line 83: to be honest if this is not needed anywhere else in the codebase I feel like the result_manager_factory could be enough without having to go through a list (but can be done in another PR)

anta/models.py
```diff
@@ -448,15 +448,16 @@ def __init__(
...
         self._init_inputs(inputs)
-        if self.result.result == AntaTestStatus.UNSET:
+        if hasattr(self, "inputs"):
             self._init_commands(eos_data)
+            if res_ow := self.inputs.result_overwrite:
+                if res_ow.categories:
+                    self.result.categories = res_ow.categories
+                if res_ow.description:
+                    self.result.description = res_ow.description
+                self.result.custom_field = res_ow.custom_field
```
gmuloc, line 460: why is this moving? this should be part of the inputs still no?

anta/models.py
```diff
@@ -448,15 +448,16 @@ def __init__(
...
-        self.result: TestResult = TestResult(
-            name=device.name,
-            test=self.name,
-            categories=self.categories,
-            description=self.description,
-        )
+        self.result: TestResult = TestResult(name=device.name, test=self.name, categories=self.categories, description=self.description)
         self._init_inputs(inputs)
-        if self.result.result == AntaTestStatus.UNSET:
+        if hasattr(self, "inputs"):
```
gmuloc, line 453: this looks breaking - if there is no `inputs` then we don't init the command anymore?

anta/models.py
```diff
@@ -461,6 +461,7 @@ def _init_inputs(self, inputs: dict[str, Any] | AntaTest.Input | None) -> None:
             if res_ow.description:
                 self.result.description = res_ow.description
             self.result.custom_field = res_ow.custom_field
+        self.result.inputs = self.inputs
```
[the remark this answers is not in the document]
mtache: No, by default pydantic does not validate on field assignments.
Trying to see if I can improve the benchmarks but I'm close to affirm that this is the cost to have inputs in the results.

anta/reporter/__init__.py
```diff
@@ -102,35 +103,84 @@ def _color_result(self, status: AntaTestStatus) -> str:
...
+        def add_result(result: TestResult) -> None:
+            add_line(result)
+            for index, atomic_res in enumerate(result.atomic_results):
+                add_line(atomic_res, f"{index+1}/{len(result.atomic_results)}")
+
+        for result in manager.results_by_category:
+            add_result(result)
+        return table
+
+    def report(self, manager: ResultManager) -> Table:
```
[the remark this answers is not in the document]
mtache: This is internal to ANTA. The title argument should have never been there anyway since it's always the default value... It is never used.

anta/models.py
mtache: This is not breaking.
The first condition was to verify that `self.result.result` has its default value (UNSET) and has not been set to ERROR because of input validation error.
I find the condition `hasattr(self, "inputs")` more readable and completely equivalent.

anta/models.py
mtache: It is equivalent.
It is overriding the TestResult fields with some inputs.
Having move outside of "init_inputs" is more readable since it is not initiating any inputs, but using initiated inputs to modify the result.

anta/result_manager/models.py
mtache: Do you have an example where this could be needed?

anta/result_manager/models.py
mtache: Yes it should be. But it causes circular import.
See https://github.com/aristanetworks/anta/issues/994, to implement in another PR.

anta/cli/nrfu/utils.py
```diff
@@ -80,7 +80,7 @@ def print_settings(
     console.print()
 
 
-def print_table(ctx: click.Context, group_by: Literal["device", "test"] | None = None) -> None:
+def print_table(ctx: click.Context, expand_atomic: bool, group_by: Literal["device", "test"] | None) -> None:
```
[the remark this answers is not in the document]
mtache: I was not doing this because of https://docs.astral.sh/ruff/rules/boolean-default-value-positional-argument/ but let's ignore Ruff...

tests/units/anta_tests/__init__.py
mtache: This has been cleaned as part of the benchmark PR, and there is a some check there to be removed: https://github.com/aristanetworks/anta/blob/main/tests/benchmark/test_anta.py#L82

https://github.com/aristanetworks/anta/issues/996

tests/units/anta_tests/__init__.py
mtache: https://github.com/aristanetworks/anta/issues/995

anta/cli/nrfu/utils.py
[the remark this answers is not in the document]
mtache: As per our discussion, this is not public API.

anta/reporter/__init__.py
```diff
@@ -102,35 +105,88 @@ def _color_result(self, status: AntaTestStatus) -> str:
...
+            if isinstance(result, TestResult):
+                categories = ", ".join(convert_categories(result.categories))
+                device = str(result.name)
+                test = result.test
+            else:
+                test = name
+            state = self._color_result(result.result)
+            message = self._split_list_to_txt_list(result.messages) if len(result.messages) > 0 else ""
+            inputs = (
+                safe_dump(safe_load(result.inputs.model_dump_json(exclude_none=True)), indent=2, width=math.inf)
```
carl-baillargeon, line 140: safe_dump + safe_load is pretty slow. On a medium-large catalog:
`AntaCatalog.yaml()`:
image
`AntaCatalog.to_json()`:
image

2.33s vs 0.01s :P

anta/cli/debug/utils.py
```diff
@@ -42,13 +42,11 @@ def debug_options(f: Callable[..., Any]) -> Callable[..., Any]:
...
         ctx: click.Context,
-        *args: tuple[Any],
+        *args: Any,  # noqa: ANN401
         inventory: AntaInventory,
         device: str,
-        **kwargs: Any,
-    ) -> Any:
-        # TODO: @gmuloc - tags come from context https://github.com/aristanetworks/anta/issues/584
-        # ruff: noqa: ARG001
+        **kwargs: Any,  # noqa: ANN401
```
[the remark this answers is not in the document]
mtache: Because of this: https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

anta/reporter/__init__.py
mtache: Yeah, it's a copy/paste from AntaCatalog method indeed, so nothing new on this PR.
As mentionned there, it is related to this issue: https://github.com/pydantic/pydantic/issues/1043

I opened this to remember to work on that: https://github.com/aristanetworks/anta/issues/998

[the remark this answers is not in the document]
mtache: Prevent the YAML library to add new lines in the dump, which is what we want when dumping a file.
Not sure what value to use here.

anta/tests/connectivity.py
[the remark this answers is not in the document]
mtache: I do not see how to implement that.
A reference means to have a *generic* way to point somewhere in the input structure of the parent. Like JSONPath in JSON or XPath in XML.
I do not see such type in pydantic, neither the way to create one.

docs/contribution.md
mtache: No unit test case expects unset (test has never run) or error (an unexpected error has been caught by the ANTA framework, not by the test code).

anta/tests/connectivity.py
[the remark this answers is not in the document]
mtache: Let's do like we discussed: add options to show inputs of the parent, of the atomic results or all of them.

anta/reporter/__init__.py
```diff
@@ -102,35 +105,88 @@ def _color_result(self, status: AntaTestStatus) -> str:
...
+        manager
+            A ResultManager instance.
+
+        Returns
+        -------
+        Table
+            A fully populated rich `Table`.
+        """
+        table = Table(title=ReportTable.TITLE_ALL, show_lines=True)
+        headers = ["Category", "Device", "Test", "Status", "Message(s)"]
```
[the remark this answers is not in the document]
mtache: I moved that to the expanded view only as it is redundant/less readable to have the same sentence many times in the table.

tests/benchmark/test_runner.py
mtache: Good catch

tests/units/reporter/test__init__.py
mtache: Implemented in test_titles below

anta/tests/connectivity.py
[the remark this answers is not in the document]
mtache: Done

.vscode/settings.json
```diff
@@ -10,5 +10,7 @@
   "pylint.args": [
     "--rcfile=pyproject.toml"
   ],
+  "python-envs.defaultEnvManager": "ms-python.python:pyenv",
+  "python-envs.pythonProjects": [],
```
carl-baillargeon: Who uses pyenv these days :D uv for the win!

anta/cli/nrfu/utils.py
```diff
@@ -116,16 +124,18 @@ def print_json(ctx: click.Context, output: pathlib.Path | None = None) -> None:
             ctx.exit(ExitCode.USAGE_ERROR)
 
 
-def print_text(ctx: click.Context) -> None:
+def print_text(ctx: click.Context, *, expand_atomic: bool) -> None:
```
carl-baillargeon: Similar to my other comment. I vote for `expand: bool`.

anta/reporter/__init__.py
```diff
@@ -136,9 +137,9 @@ def generate(self, manager: ResultManager) -> Table:
                 - `columns.category`
                 - `columns.device`
                 - `columns.test`
+                - `columns.description`
```
carl-baillargeon: This is not following the order in the column list below.

anta/cli/nrfu/commands.py
```diff
@@ -35,10 +35,19 @@
...
+@click.option(
+    "--expand",
+    "-x",
+    default=False,
+    show_envvar=True,
+    is_flag=True,
+    show_default=True,
+    help="Flag to indicate if test descriptions and atomic results should be shown.",
+)
+def table(ctx: click.Context, *, group_by: Literal["device", "test"] | None, sort_by: tuple[str] | None, expand: bool) -> None:
```
carl-baillargeon: Should we update the `sort_by` option to accept `atomic_results`? I would say no but technically this is now supported by `ResultManager.sort()` and `ResultManager.get_results()`.

carl-baillargeon (commented):
Few minor comments but overall all my tests were successful.

Atomic results are also rendered properly in the JSON report of anta_runner:
```
    {
        "name": "DC1-LEAF1A",
        "test": "VerifyReachability",
        "categories": [
            "connectivity"
        ],
        "description": "Verifies reachability to BGP neighbors with an update source configured.",
        "result": "success",
        "messages": [],
        "atomic_results": [
            {
                "description": "Destination 10.1.0.1 from Loopback0 in VRF default",
                "result": "success",
                "messages": []
            },
            {
                "description": "Destination 10.1.0.2 from Loopback0 in VRF default",
                "result": "success",
                "messages": []
            }
        ],
        "custom_field": "Generated by AVD"
    },
```

In dry-run, we don't see the atomic results but I don't think we can do anything about it since they are added inside the coroutines which we never await in dry-run. I think that's fine for now though. Later when/if we add the inputs we'll have more context in the dry-run.

anta/reporter/__init__.py
```diff
@@ -161,6 +162,62 @@ def generate(self, manager: ResultManager) -> Table:
...
+
+            if isinstance(result, TestResult):
+                categories = ", ".join(convert_categories(result.categories))
+                device = str(result.name)
+                test = result.test
+            elif suffix is not None:
+                test = f"{result.parent.test} {suffix}"
+
+            state = self._color_result(result.result)
+            message = self._split_list_to_txt_list(result.messages) if len(result.messages) > 0 else ""
```
carl-baillargeon: Not sure if we should have the parent entry message in the table when expand is enabled. It seems redundant to me:
image

We see the same failure message twice. Maybe we could have a generic message like "See atomic results below for details" or something in the parent entry.

anta/result_manager/models.py
```diff
@@ -89,6 +89,60 @@ def is_error(self, message: str | None = None) -> None:
...
+    messages : list[str]
+        Messages reported by the test.
+    """
+
+    _parent: TestResult
+    description: str
+    result: AntaTestStatus = AntaTestStatus.UNSET
+    messages: list[str] = []
+
+    def __init__(self, **data: Any) -> None:  # noqa: ANN401
```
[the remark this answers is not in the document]
gmuloc: leftover from me cleaning inputs ..

anta/reporter/__init__.py
carl-baillargeon: Nevermind. It wouldn't be consistent with other reports if we do that.

Issue #809, opened 2024-08-30
Extra attributes should be forbidden in nested Input models


carl-baillargeon:
Nested input models like `class BgpAfi(BaseModel)` don't inherit from `AntaTest.Input` meaning they don't forbid extra attributes. This could lead to unexpected behaviors for a user who wants to override a default value but accidently provides the wrong key.

BGP tests for example. If a user provides `vrfs` instead of `vrf`, ANTA will silently ignore `vrfs` and use the default value of `vrf`. An error should be raised to the user since `vrfs` is not supported.

carl-baillargeon:
Nested input models are being moved over to `anta.input_models` package and `model_config = ConfigDict(extra="forbid")` should be added at the same time.

carl-baillargeon:
Completed as part of #388

Issue #587, opened 2024-03-18
Have nicer result failure messages


carl-baillargeon:
Would be nice to have a more human readable format for the test result failures messages:
![image](https://github.com/arista-netdevops-community/anta/assets/63206086/bd3d551d-f910-4e95-a7c5-6ecd462f93d9)

BGP tests are also not very user friendly:
![image](https://github.com/arista-netdevops-community/anta/assets/63206086/da408c18-6b1e-4cdb-a92c-0d6a2f4fd7da)

Could be related to #427

carl-baillargeon:
Adding VerifySnmpContact to the list:
![image](https://github.com/arista-netdevops-community/anta/assets/63206086/e7c5b89d-8360-4371-b208-79bd9439d5f7)

carl-baillargeon:
Some tests failure messages will be improved as part of #388

carl-baillargeon:
@vitthalmagadum and @geetanjalimanegslab Please investigate remaining tests so we know how big of an effort it would be to remove the `[ ]` and `{ }` in the failure messages.

Thanks

geetanjalimanegslab:
Hi @carl-baillargeon 
Here is the list of test files and the number of messages need to be updated:

**Test files needs to be updated are:**
**1. test_generic.py 
2. test_isis.py
3. test_ospf.py
4. test_aaa.py
5. test_avt.py
6. test_bfd.py
7.  test_connectivity.py
8. test_field_notices.py
9. test_hardware.py
10. test_interfaces.py
11. test_logging.py
12. test_mlag.py
13. test_ptp.py
14.  test_security.py
15. test_services.py
16. test_snmp.py
17. test_software.py
18. test_stp.py
19. test_system.py
20.  test_vxlan.py**

Issue #388, opened 2023-09-12
Enhancement: Input subclasses refactoring


gmuloc:
Tests Input subclasses should be moved outside of test classes, in other modules to allow reusability across tests and easier imports by external framework leveraging ANTA (AVD).

Solution:

Models are being moved to `anta.input_models.<module>`. Each test will keep a class variable as a reference of these models to avoid breaking changes. The references will be removed in ANTA v2.0.

Example:
```
from anta.input_models.connectivity import Host

class Input(AntaTest.Input):
    """Input model for the VerifyReachability test."""

    hosts: list[Host]
    """List of host to ping."""
    Host: ClassVar[type[Host]] = Host
```

- As part of this refactoring, `model_config = ConfigDict(extra="forbid")` should be added to these models to fix #809. 
- A `__str__`  should also be added to the models and be leveraged to improve failure messages, to fix #587.
- Need to add documentation for input models. Ref #912 

# Checklist (in priority):

High priority (AVD tests or high reusability):
- [x] VerifyReachability.Input.Host #912
- [x] VerifyInterfacesStatus.Input.InterfaceState #899
- [x] VerifyNTPAssociations.Input.NTPServer #902
- [x] VerifyLLDPNeighbors.Input.Neighbor #915 
- [x] VerifySpecificIPSecConn.Input.IPSecPeers & IPSecConn - https://github.com/aristanetworks/anta/pull/933
- [x] VerifyStunClient..Input.ClientAddress - https://github.com/aristanetworks/anta/pull/934
- [x] VerifyAVTSpecificPath.Input.AVTPaths  -  https://github.com/aristanetworks/anta/pull/939
- [x] VerifyDNSServers.Input.DnsServer https://github.com/aristanetworks/anta/pull/920

The following 3 BGP tests should use the same model to replace <VerifyBGP...>.Input.BgpAfi:
- [x] VerifyBGPPeerCount, VerifyBGPPeersHealth, VerifyBGPSpecificPeers #888

These BGP tests should use the same `BgpPeer` model:
- [x] VerifyBGPPeerMPCaps #914
- [x] VerifyBGPPeerASNCap #914
- [x] VerifyBGPPeerRouteRefreshCap #914
- [x] VerifyBGPPeerMD5Auth #914
- [x] VerifyBGPAdvCommunities #914
- [x] VerifyBGPTimers #914
- [x] VerifyBGPPeerDropStats #914
- [x] VerifyBGPPeerUpdateErrors #914
[...]

PR #955, opened 2024-12-09
fix(anta.tests): Cleaning up Services tests module (VerifyErrdisableRecovery) 

vitthalmagadum:
# Description

Refactoring Services tests module to address the following issues:

- https://github.com/aristanetworks/anta/issues/809
- https://github.com/aristanetworks/anta/issues/587
- https://github.com/aristanetworks/anta/issues/388

# Checklist:


anta/tests/services.py
```diff
@@ -166,12 +163,24 @@ def test(self) -> None:
...
 
 class VerifyErrdisableRecovery(AntaTest):
-    """Verifies the errdisable recovery reason, status, and interval.
+    """Verifies the errdisable recovery reason.
+
+    This test performs the following checks for each specified Errdisable reason:
+
+      1. Checking that the errdisable recovery reason exists.
+      2. Verifying that the status of the errdisable recovery reason is set as specified.
+      3. Ensuring the interval value matches the expected input.
```
carl-baillargeon, line 172:
```suggestion
    This test performs the following checks for each specified error disable reason:

      1. Verifying if the specified error disable reason exists.
      2. Checking if the recovery timer status matches the expected enabled/disabled state.
      3. Validating that the timer interval matches the configured value.
```

anta/tests/services.py
```diff
@@ -166,12 +163,24 @@ def test(self) -> None:
...
-    * Success: The test will pass if the errdisable recovery reason status is enabled and the interval matches the input.
-    * Failure: The test will fail if the errdisable recovery reason is not found, the status is not enabled, or the interval does not match the input.
+    * Success: The test will pass if:
+        - The specified errdisable recovery reason is present in the configuration.
+        - The status of the recovery reason is set as specified.
+        - The configured interval matches the input value.
+    * Failure: The test will fail if:
+        - The errdisable recovery reason is not found in the configuration.
+        - The status of the recovery reason is set to "Disabled" or any unexpected value.
+        - The interval value does not match the expected input.
```
carl-baillargeon, line 183:
```suggestion
    * Success: The test will pass if:
        - The specified error disable reason exists.
        - The recovery timer status matches the expected state.
        - The timer interval matches the configured value.
    * Failure: The test will fail if:
        - The specified error disable reason does not exist.
        - The recovery timer status does not match the expected state.
        - The timer interval does not match the configured value.
```

anta/tests/services.py
```diff
@@ -195,42 +206,28 @@ class Input(AntaTest.Input):
...
-                break
-
-            if not reason_found:
-                self.result.is_failure(f"`{input_reason}`: Not found.\n")
+            if not (reason_output := get_item(errdisable_reasons, "reason", error_reason.reason)):
+                self.result.is_failure(f"{error_reason} - Not found")
+                continue
+
+            if not all([error_reason.status == (act_status := reason_output["status"]), error_reason.interval == (act_interval := int(reason_output["interval"]))]):
+                self.result.is_failure(f"{error_reason} - Incorrect reason details - Status: {act_status} Interval: {act_interval}")
```
carl-baillargeon, line 233:
```suggestion
                self.result.is_failure(f"{error_reason} - Incorrect configuration - Status: {act_status} Interval: {act_interval}")
```

carl-baillargeon (changes requested):

anta/input_models/services.py
```diff
@@ -29,3 +32,24 @@ def __str__(self) -> str:
...
         return f"Server {self.server_address} (VRF: {self.vrf}, Priority: {self.priority})"
+
+
+class ErrDisableReason(BaseModel):
+    """Model for an errdisable reason."""
+
+    model_config = ConfigDict(extra="forbid")
+    reason: ErrDisableReasons
+    """Type or name of the errdisable reason."""
+    status: Literal["Enabled", "Disabled"] = "Enabled"
```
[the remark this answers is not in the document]
vitthalmagadum: In the test, we previously checked that the status was enabled as static. Implementing this change (enabled: False) might introduce a breaking behavior.
@MaheshGSLAB @carl-baillargeon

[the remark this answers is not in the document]
carl-baillargeon: Ah yes I missed that. In that case, to avoid confusion let's keep the existing behaviour (no status/enabled knob) and we just check that it is "Enabled" in the code. If we get the request to explicitly check if recovery is "Disabled" for a reason later on, we'll add it. I doubt since by default they are all disabled in EOS.

anta/input_models/services.py
```diff
@@ -29,3 +32,24 @@ def __str__(self) -> str:
...
+class ErrdisableRecovery(BaseModel):
+    """Model for the error disable recovery functionality."""
+
+    model_config = ConfigDict(extra="forbid")
+    reason: ErrdisableReason
+    """Name of the error disable reason."""
+    status: Literal["Enabled", "Disabled"] = "Enabled"
+    """Operational status of the reason. Defaults to 'Enabled'."""
+    timer_interval: int = Field(ge=30, le=86400)
+    """Timer interval of the reason in seconds. Required field in the `VerifyErrdisableRecovery` test."""
```
carl-baillargeon, line 46:
```suggestion
    """Timer interval of the reason in seconds."""
```
This field is not optional so no need to say it's required.

anta/input_models/services.py
```diff
@@ -29,3 +32,24 @@ def __str__(self) -> str:
...
+
+class ErrdisableRecovery(BaseModel):
+    """Model for the error disable recovery functionality."""
+
+    model_config = ConfigDict(extra="forbid")
+    reason: ErrdisableReason
+    """Name of the error disable reason."""
+    status: Literal["Enabled", "Disabled"] = "Enabled"
+    """Operational status of the reason. Defaults to 'Enabled'."""
+    timer_interval: int = Field(ge=30, le=86400)
```
carl-baillargeon, line 45: Apologies, we should keep this as `interval` to avoid breaking changes.

anta/tests/services.py
```diff
@@ -193,44 +204,35 @@ class VerifyErrdisableRecovery(AntaTest):
...
         """List of errdisable reasons."""
-
-        class ErrDisableReason(BaseModel):
-            """Model for an errdisable reason."""
-
-            reason: ErrDisableReasons
-            """Type or name of the errdisable reason."""
-            interval: ErrDisableInterval
-            """Interval of the reason in seconds."""
+        ErrDisableReason: ClassVar[type[ErrdisableRecovery]] = ErrdisableRecovery
```
carl-baillargeon, line 209:
```suggestion
        ErrDisableReason: ClassVar[type[ErrdisableRecovery]] = ErrDisableReason
```
Since we are chaning the name, we should keep `ErrDisableReason` in the input models with a deprecation, just like we do with other models (to be removed in ANTA v.2.0).

vitthalmagadum: Updated. Thanks!!

anta/input_models/services.py
vitthalmagadum: Updated

anta/input_models/services.py
vitthalmagadum: Updated

carl-baillargeon (approved):
LGTM if CI passes.

PR #977, opened 2024-12-23
fix(anta.tests): Fix wrong type used for inputs for VerifyInterfaceIPv4

### LINKED ISSUE #973: bug: VerifyInterfaceIPv4 seems to use wrong type for IPv4 Addresses
Hi,
when running an anta test like the following via CLI:
```---
anta.tests.interfaces:
  - VerifyInterfaceIPv4:
      interfaces:
        - name: Ethernet3
          primary_ip: 10.20.2.2/30
      filters:
        tags: ["leaf1"]
      result_overwrite:
        description: "Underlay/IPConfig"
```

it get an error like this:
```
CRITICAL Test catalog is invalid! (from anta/anta_dev_test.yml)                                                                                                        logger.py:121
                    ValidationError: 1 validation error for AntaCatalogFile                                                                                                                    
                    inputs                                                                                                                                                                     
                      VerifyInterfaceIPv4 test inputs are not valid: 6 validation errors for Input                                                                                             
                            interfaces.0.primary_ip                                                                                                                                            
                              Input is not a valid IPv4 network
```
The code for anta test `VerifyInterfaceIPv4` expects an instance of `ipaddress.IPv4Network` for `primary_ip` and `secondary_ip`. However the test is intended to validate host addresses, yet `ipaddress.IPvNetwork` will raise an expection by default when initialized with an address where host bits are set. So when providing a v4 address in CIDR notation in the catalog the above error is raised. 
Since the unit test for the anta test uses the first addresses of /31 networks this error is not detected (for /31 networks the network address equals the first host address - therefore no exception occurs). You should be able to reproduce the issue by just updating one of the inputs in the unit test. For example change "primary_ip": "172.30.11.0/31" to  "primary_ip": "172.30.11.1/31".

The issue might be fixed by using `ipaddress.IPv4Interface` instead of `ipaddress.IPv4Network` for this anta test i guess.

gmuloc:
# Description

Use correct type for inputs

Fixes #973

# Checklist:

### DIFF
--- a/anta/tests/interfaces.py
+++ b/anta/tests/interfaces.py
@@ -8,7 +8,7 @@
 from __future__ import annotations
 
 import re
-from ipaddress import IPv4Network
+from ipaddress import IPv4Interface
 from typing import Any, ClassVar
 
 from pydantic import BaseModel, Field
@@ -629,9 +629,9 @@ class VerifyInterfaceIPv4(AntaTest):
       - VerifyInterfaceIPv4:
           interfaces:
             - name: Ethernet2
-              primary_ip: 172.30.11.0/31
+              primary_ip: 172.30.11.1/31
               secondary_ips:
-                - 10.10.10.0/31
+                - 10.10.10.1/31
                 - 10.10.10.10/31
     ```
     """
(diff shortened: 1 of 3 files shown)


carl-baillargeon (approved):
LGTM

PR #979, opened 2024-12-24
ci: Add environment to protect publishing

gmuloc:
# Description

ci stuff for pypi

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,11 @@ jobs:
   pypi:
     name: Publish version to Pypi servers
     runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/anta
+    permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -9,6 +9,11 @@ jobs:
   pypi:
     name: Publish version to Pypi servers
     runs-on: ubuntu-latest
+    environment:
+      name: pypi
```
gmuloc, line 13:
```suggestion
      name: production
```

mtache (approved):

.github/workflows/release.yml
```diff
@@ -9,6 +9,11 @@ jobs:
   pypi:
     name: Publish version to Pypi servers
```
mtache, line 10:
```suggestion
    name: Publish Python 🐍 distribution 📦 to PyPI
```

PR #981, opened 2024-12-24
test: idempotency of setup logger

gmuloc:
# Description

Great news everyone - it is now MUCH faster to run the tests.. Locally it means 9s vs 74s for my machine.

The issue was that we were adding too many RichHandler when calling CliRunner multiple times...

**NOTE:** Logging need some more love.

# Checklist:

### DIFF
--- a/anta/logger.py
+++ b/anta/logger.py
@@ -72,7 +72,7 @@ def setup_logging(level: LogLevel = Log.INFO, file: Path | None = None) -> None:
     # Add RichHandler for stdout if not already present
     maybe_add_rich_handler = True
     if root.hasHandlers():
-        maybe_add_rich_handler = any(handler.get_name() == "ANTA_RICH_HANDLER" for handler in root.handlers)
+        maybe_add_rich_handler = all(handler.get_name() != "ANTA_RICH_HANDLER" for handler in root.handlers)
 
     if maybe_add_rich_handler:
         rich_handler = RichHandler(markup=True, rich_tracebacks=True, tracebacks_show_locals=False)
--- a/anta/logger.py
+++ b/anta/logger.py
@@ -102,7 +102,7 @@ def _get_rich_handler(logger_instance: logging.Logger) -> logging.Handler | None
 
 def _maybe_add_rich_handler(loglevel: int, logger_instance: logging.Logger) -> None:
     """Add RichHandler for stdout if not already present."""
-    if _get_rich_handler is not None:
+    if _get_rich_handler(logger_instance) is not None:
         # Nothing to do.
         return
 
--- a/anta/logger.py
+++ b/anta/logger.py
@@ -88,6 +88,10 @@ def setup_logging(level: LogLevel = Log.INFO, file: Path | None = None) -> None:
 
 def _get_file_handler(logger_instance: logging.Logger, file: Path) -> logging.FileHandler | None:
     """Return the FileHandler if present."""
+    for handler in logger_instance.handlers:
+        if isinstance(handler, logging.FileHandler):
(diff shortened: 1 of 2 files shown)


mtache (changes requested):
Maybe do the same logic for the FileHandler?

gmuloc:
> Maybe do the same logic for the FileHandler?

maybe don't call the funciton multiple time with the same file? :p

anta/logger.py
```diff
@@ -69,13 +69,20 @@ def setup_logging(level: LogLevel = Log.INFO, file: Path | None = None) -> None:
...
 
-    # Add RichHandler for stdout
-    rich_handler = RichHandler(markup=True, rich_tracebacks=True, tracebacks_show_locals=False)
-    # Show Python module in stdout at DEBUG level
-    fmt_string = "[grey58]\\[%(name)s][/grey58] %(message)s" if loglevel == logging.DEBUG else "%(message)s"
-    formatter = logging.Formatter(fmt=fmt_string, datefmt="[%X]")
-    rich_handler.setFormatter(formatter)
-    root.addHandler(rich_handler)
+    # Add RichHandler for stdout if not already present
+    maybe_add_rich_handler = True
```
carl-baillargeon, line 73:
```suggestion
    needs_rich_handler = True
```

carl-baillargeon (approved):
Minor suggestion on variable name otherwise LGTM, tested locally:
image

Thank you for this, a lot faster now!!

mtache:
> > Maybe do the same logic for the FileHandler?
> 
> maybe don't call the function multiple times with the same file? :p

This point of view is also valid for RichHandler, nevertheless we are making it idempotent.
I think this function should behave in a similar way for RichHandler and FileHandler (and it should have been idempotent from day 1).

From our unit test issue, if I write unit tests with a file as env var and I use CliRunner, I would log multiple times in the same file.

gmuloc:
> > > Maybe do the same logic for the FileHandler?
> > 
> > 
> > maybe don't call the function multiple times with the same file? :p
> 
> This point of view is also valid for RichHandler, nevertheless we are making it idempotent. I think this function should behave in a similar way for RichHandler and FileHandler (and it should have been idempotent from day 1).

Agreed

> 
> From our unit test issue, if I write unit tests with a file as env var and I use CliRunner, I would log multiple times in the same file.

Thats really a unit test problem

anta/logger.py
```diff
@@ -69,27 +69,52 @@ def setup_logging(level: LogLevel = Log.INFO, file: Path | None = None) -> None:
...
-    fmt_string = "[grey58]\\[%(name)s][/grey58] %(message)s" if loglevel == logging.DEBUG else "%(message)s"
-    formatter = logging.Formatter(fmt=fmt_string, datefmt="[%X]")
-    rich_handler.setFormatter(formatter)
-    root.addHandler(rich_handler)
-    # Add FileHandler if file is provided
-    if file:
+    # Add RichHandler for stdout if not already present
+    _maybe_add_rich_handler(loglevel, root)
+
+    # Add FileHandler if file is provided and same File Handler is not already present
```
gmuloc, line 75:
```suggestion
    # Add FileHandler if file is provided and same FileHandler is not already present
```

mtache:
NB for later:
we should not modify the root logger but an anta logger. Right way to implement that would be to delete all handlers of the anta logger at the beginning of the function.
We used root logger to have logs from other libraries, maybe we could modify specific loggers (httpx, asyncio) instead when in ANTA_DEBUG mode.

mtache (approved):

PR #1016, opened 2025-01-21
doc: Enable symbol

gmuloc:
To be discussed:

* add symbols in doc: maybe should not be added for tests
    * discussed with @mtache  - symbol for API not for tests.
* simplify mkdocstinrg overwrites (but we end up with duplicate info for inputs and we cannot have the summary table without the attribute fields)
    * discussed -> no change pour le moment


docs/api/tests/routing.bgp.md
```diff
@@ -20,31 +21,34 @@ anta_title: ANTA catalog for BGP tests
...
     options:
-      show_root_heading: false
-      show_root_toc_entry: false
-      show_bases: false
       anta_hide_test_module_description: true
-      merge_init_into_class: false
-      show_labels: true
       filters:
-        - "!^__init__"
         - "!^__str__"
```
gmuloc, line 44:
```suggestion
        - "!^__init__"
        - "!^__str__"
```

PR #1067, opened 2025-03-06
Feat(anta.tests): Add test for Maintenance mode

emilarista:
# Description

Test to verify if one or more units are under or entering maintenance mode for different reasons:

* On-boot configuration
* Interface traffic violation
* quiesce explicitly configured

# Checklist:

### DIFF
--- a/anta/tests/system.py
+++ b/anta/tests/system.py
@@ -345,3 +345,65 @@ class VerifyNTPAssociations(AntaTest):
 
             if act_condition != exp_condition or act_stratum != exp_stratum:
                 self.result.is_failure(f"{ntp_server} - Bad association - Condition: {act_condition}, Stratum: {act_stratum}")
+
+
+class VerifyMaintenance(AntaTest):
+    """Verifies that the device is not currently under maintenance.
+
+    Expected Results
+    ----------------
+    * Success: The test will pass if the device is not under or entering maintenance.
+    * Failure: The test will fail if the device is under or entering maintenance.
+
+    Examples
+    --------
+    ```yaml
+    anta.tests.system:
+      - VerifyMaintenance:
+    ```
+    """
+
+    categories: ClassVar[list[str]] = ["Maintenance"]
+    commands: ClassVar[list[AntaCommand | AntaTemplate]] = [AntaCommand(command="show maintenance", revision=1)]
+
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """Main test function for VerifyMaintenance."""
+        self.result.is_success()
+
+        # If units is not empty we have to examine the output for details.
+        if (units := get_value(self.instance_commands[0].json_output, "units")):
+            unitsundermaintenance = []
+            unitsenteringmaintenance = []
+            under = False
+            entering = False
+            causes = []
+            # Iterate over units, check for units under or entering maintenance, and examine the causes.
+            for unit, info in units.items():
+                if info["state"] == "underMaintenance":
+                    unitsundermaintenance.append(unit)
+                    under = True
+                elif info["state"] == "maintenanceModeEnter":
+                    unitsenteringmaintenance.append(unit)
+                    entering = True
+                if info["adminState"] == "underMaintenance":
+                    causes.append("quiesce is configured")
+                if info["onBootMaintenance"]:
+                    causes.append("On-boot maintenance is configured")
+                if info["intfsViolatingTrafficThreshold"]:
+                    causes.append("Interface traffic threshold violation")
+
+            # This can occur if maintenance is configured but no unit is configured with 'quiesce'.
+            if not under and not entering and not causes:
+                self.result.is_success()
+
+            # Building the error message.
+            else:
+                message = "Some units are under or entering maintenance."
+                if under:
+                    message += f" The following units are currently under maintenance: '{unitsundermaintenance}'."
+                if entering:
+                    message += f" The following units are currently entering maintenance: '{unitsenteringmaintenance}'."
+                if len(causes) > 0:
+                    message += f" Possible causes: '{causes}'"
+                self.result.is_failure(message)
--- a/anta/tests/system.py
+++ b/anta/tests/system.py
@@ -372,7 +372,7 @@ class VerifyMaintenance(AntaTest):
         self.result.is_success()
 
         # If units is not empty we have to examine the output for details.
-        if (units := get_value(self.instance_commands[0].json_output, "units")):
+        if units := get_value(self.instance_commands[0].json_output, "units"):
             unitsundermaintenance = []
             unitsenteringmaintenance = []
             under = False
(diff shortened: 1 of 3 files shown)


anta/tests/system.py
```diff
@@ -345,3 +345,65 @@ def test(self) -> None:
...
+                if info["adminState"] == "underMaintenance":
+                    causes.append("quiesce is configured")
+                if info["onBootMaintenance"]:
+                    causes.append("On-boot maintenance is configured")
+                if info["intfsViolatingTrafficThreshold"]:
+                    causes.append("Interface traffic threshold violation")
+
+            # This can occur if maintenance is configured but no unit is configured with 'quiesce'.
+            if not under and not entering and not causes:
+                self.result.is_success()
```
gmuloc, line 398: you already set is_success at the beginning so you don't need this. you can just return

anta/tests/system.py
```diff
@@ -345,3 +345,65 @@ def test(self) -> None:
...
+            # Building the error message.
+            else:
+                message = ""
+                if under:
+                    message += f"Units under maintenance: '{unitsundermaintenance}'. "
+                if entering:
+                    message += f"Units entering maintenance: '{unitsenteringmaintenance}'. "
+                if len(causes) > 0:
+                    message += f"Possible causes: '{causes}'"
+                self.result.is_failure(message)
```
gmuloc, line 409: we have a list os messages so yuo can do instead:
```suggestion
                if units_under_maintenance:
                    self.result.is_failure(f"Units under maintenance: '{', '.join(units_under_maintenance)}'.")
                if units_entering_maintenance:
                    self.result.is_failure(f"Units entering maintenance: '{', '.join(units_entering_maintenance)}'.")
                if len(causes) > 0:
                    self.result.is_failure(f"Possible causes: '{', '.join(causes}'")
```

anta/tests/system.py
```diff
@@ -345,3 +345,65 @@ def test(self) -> None:
...
+    def test(self) -> None:
+        """Main test function for VerifyMaintenance."""
+        self.result.is_success()
+
+        # If units is not empty we have to examine the output for details.
+        if units := get_value(self.instance_commands[0].json_output, "units"):
+            unitsundermaintenance = []
+            unitsenteringmaintenance = []
+            under = False
+            entering = False
```
gmuloc, line 379: you don't need the flags as you just need to check if the lists are empty or not

anta/tests/system.py
```diff
@@ -345,3 +345,65 @@ def test(self) -> None:
...
+
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """Main test function for VerifyMaintenance."""
+        self.result.is_success()
+
+        # If units is not empty we have to examine the output for details.
+        if units := get_value(self.instance_commands[0].json_output, "units"):
+            unitsundermaintenance = []
+            unitsenteringmaintenance = []
```
gmuloc, line 377:
```suggestion
            units_under_maintenance = []
            units_entering_maintenance = []
```

anta/tests/system.py
```diff
@@ -345,3 +345,65 @@ def test(self) -> None:
...
+            under = False
+            entering = False
+            causes = []
+            # Iterate over units, check for units under or entering maintenance, and examine the causes.
+            for unit, info in units.items():
+                if info["state"] == "underMaintenance":
+                    unitsundermaintenance.append(unit)
+                    under = True
```
gmuloc, line 385:
```suggestion
                    units_under_maintenance.append(unit)
```

anta/tests/system.py
```diff
@@ -345,3 +345,65 @@ def test(self) -> None:
...
+                elif info["state"] == "maintenanceModeEnter":
+                    unitsenteringmaintenance.append(unit)
+                    entering = True
```
gmuloc, line 388:
```suggestion
                    units_entering_maintenance.append(unit)
```

anta/tests/system.py
```diff
@@ -345,3 +345,65 @@ def test(self) -> None:
...
+                if info["adminState"] == "underMaintenance":
+                    causes.append("quiesce is configured")
+                if info["onBootMaintenance"]:
+                    causes.append("On-boot maintenance is configured")
+                if info["intfsViolatingTrafficThreshold"]:
+                    causes.append("Interface traffic threshold violation")
```
gmuloc, line 394:
```suggestion
                    causes.append("quiesce is configured.")
                if info["onBootMaintenance"]:
                    causes.append("On-boot maintenance is configured.")
                if info["intfsViolatingTrafficThreshold"]:
                    causes.append("Interface traffic threshold violation.")
```

anta/tests/system.py
emilarista: True, thanks!

anta/tests/system.py
emilarista: Cool, have implemented this, thanks!

anta/tests/system.py
emilarista: Fixed!

anta/tests/system.py
emilarista: Fixed!

anta/tests/system.py
emilarista: Fixed!

anta/tests/system.py
gmuloc: yeah no I did this in the wrong order
ignore this one :)

anta/tests/system.py
```diff
@@ -345,3 +345,55 @@ def test(self) -> None:
...
+        self.result.is_success()
+
+        # If units is not empty we have to examine the output for details.
+        if not (units := get_value(self.instance_commands[0].json_output, "units")):
+            return
+        units_under_maintenance = [unit for unit, info in units.items() if info["state"] == "underMaintenance"]
+        units_entering_maintenance = [unit for unit, info in units.items() if info["state"] == "maintenanceModeEnter"]
+        causes = set()
+        # Iterate over units, check for units under or entering maintenance, and examine the causes.
+        for info in units.values():
```
gmuloc, line 381: so here we end up iterating 3 times over the list which is not the most efficient. (I know why you made the changes like this but for the sake of efficiency we may prefer to add back the two ifs from line 377 and 378 inside the for loop - anyways it is the same complexity in theory.

anta/tests/system.py
```diff
@@ -345,3 +345,55 @@ def test(self) -> None:
...
+            if info["adminState"] == "underMaintenance":
+                causes.add("Quiesce is configured")
+            if info["onBootMaintenance"]:
+                causes.add("On-boot maintenance is configured")
+            if info["intfsViolatingTrafficThreshold"]:
+                causes.add("Interface traffic threshold violation")
+
+        # Declare success if maintenance is configured but no unit is configured with 'quiesce'.
+        if not units_under_maintenance and not units_entering_maintenance and not causes:
+            return
```
gmuloc, line 391: Actually you don't need this if statement :) if there is any it will set the result to failure (less complexity yay!):

```suggestion
```

carl-baillargeon (approved):
LGTM! Thanks Emil

PR #1127, opened 2025-03-27
ci: Remove reference to anta.ninja in dockerfile

gmuloc:
# Description

cf title

`anta.ninja` it seems we did not catch dem all.


carl-baillargeon (approved):
LGTM.

PR #1133, opened 2025-04-02
feat(anta.tests): Added fix to Accept ZTP as reload cause

### LINKED ISSUE #1130: feat(anta.tests): Accept ZTP as reload cause
Currently ANTA considers a failed test when the reload cause of a device is because of ZTP.

While this is not the best practice, a reload cause of exiting Zero Touch Provisioning should not be considered as a failed test.

This is the output of the CLI command in the device:

```
DUT#show reload cause | json
{
    "resetCauses": [
        {
            "description": "System reloaded due to Zero Touch Provisioning",
            "timestamp": 1729856740.0,
            "recommendedAction": "No action necessary.",
            "debugInfoIsDir": false
        }
    ],
    "full": false
}
```

geetanjalimanegslab:
# Description
Added fix to Accept ZTP as reload cause

Fixes #1130

# Checklist:

### DIFF
--- a/anta/tests/system.py
+++ b/anta/tests/system.py
@@ -73,8 +73,8 @@ class VerifyReloadCause(AntaTest):
 
     Expected Results
     ----------------
-    * Success: The test will pass if there are NO reload causes or if the last reload was caused by the user or after an FPGA upgrade.
-    * Failure: The test will fail if the last reload was NOT caused by the user or after an FPGA upgrade.
+    * Success: The test passes if there are no reload causes, or if the last reload was user-initiated, after an FPGA upgrade, or caused by Zero Touch Provisioning.
+    * Failure: The test will fail if the last reload was NOT caused by the user or after an FPGA upgrade or due to Zero Touch Provisioning.
     * Error: The test will report an error if the reload cause is NOT available.
 
     Examples
@@ -98,10 +98,7 @@ class VerifyReloadCause(AntaTest):
             return
         reset_causes = command_output["resetCauses"]
         command_output_data = reset_causes[0].get("description")
-        if command_output_data in [
-            "Reload requested by the user.",
-            "Reload requested after FPGA upgrade",
-        ]:
+        if command_output_data in ["Reload requested by the user.", "Reload requested after FPGA upgrade", "System reloaded due to Zero Touch Provisioning"]:
             self.result.is_success()
         else:
             self.result.is_failure(f"Reload cause is: {command_output_data}")
--- a/tests/units/anta_tests/test_system.py
+++ b/tests/units/anta_tests/test_system.py
@@ -63,6 +63,25 @@ DATA: list[dict[str, Any]] = [
         "expected": {"result": "success"},
     },
     {
+        "name": "success-valid-reload-cause",
+        "test": VerifyReloadCause,
+        "eos_data": [
+            {
+                "resetCauses": [
+                    {
+                        "description": "System reloaded due to Zero Touch Provisioning",
+                        "timestamp": 1729856740.0,
+                        "recommendedAction": "No action necessary.",
+                        "debugInfoIsDir": False,
+                    }
+                ],
+                "full": False,
+            }
+        ],
+        "inputs": None,
+        "expected": {"result": "success"},
+    },
+    {
         "name": "failure",
         "test": VerifyReloadCause,
         # The failure cause is made up


vitthalmagadum (approved):

anta/tests/system.py
```diff
@@ -98,10 +98,7 @@ def test(self) -> None:
             return
         reset_causes = command_output["resetCauses"]
         command_output_data = reset_causes[0].get("description")
-        if command_output_data in [
-            "Reload requested by the user.",
-            "Reload requested after FPGA upgrade",
-        ]:
+        if command_output_data in ["Reload requested by the user.", "Reload requested after FPGA upgrade", "System reloaded due to Zero Touch Provisioning"]:
```
carl-baillargeon, line 101: We should add an optional list of causes as input with these 3 causes as default.
```
class Input(AntaTest.Input):
    allowed_causes: list[str] = Field(default=["Reload requested by the user", ...])
```

carl-baillargeon (changes requested):

anta/tests/system.py
gmuloc: if we want to be very non breaking in behavior we should add the two first causes as default and the ZTP one as an extra to accept.
we need to create an enum of causes that makes sense rather than expecting the user to type in the string.

open to suggestions here but I guess sonething like 

```
* User Request
* FPGA Upgrade
* ZTP
```

each mapping to an error message we would look at?

carl-baillargeon: Good call on the Enum. What about this:
```
class ReloadCauses(str, Enum):
    USER = "Reload requested by the user"
    FPGA = "Reload requested after FPGA upgrade"
    ZTP = "System reloaded due to Zero Touch Provisioning"
```

That looks clean but that means we must find all the possible causes right :P?

gmuloc: or add them as requested in future versions ... ? :)

geetanjalimanegslab: @carl-baillargeon  and @gmuloc I have updated the PR with the suggested changes. Please let me know if anything else needs to be updated.

Thanks!

anta/tests/system.py
```diff
@@ -73,21 +82,39 @@ class VerifyReloadCause(AntaTest):
...
     ```
     """
 
     categories: ClassVar[list[str]] = ["system"]
     commands: ClassVar[list[AntaCommand | AntaTemplate]] = [AntaCommand(command="show reload cause", revision=1)]
 
+        """Input model for the VerifyReloadCause test."""
+
+        allowed_causes: list[str] = Field(default=["USER", "FPGA"])
```
carl-baillargeon, line 105: It feels weird to not let Pydantic handle the validation here. I think we should instead use `Literal["USER", "FPGA", "ZTP"] = Field(default=["USER", "FPGA"]` and have a mapping with their corresponding messages.

We can even create a custom type called `ReloadCause` with a proper description including the corresponding messages to display properly in our documentation.

We wouldn't need the validator below and the enum in that case.

@gmuloc Thoughts?

anta/custom_types.py
```diff
@@ -187,6 +187,23 @@ def update_bgp_redistributed_proto_user(value: str) -> str:
     return value
 
 
+def convert_reload_cause(value: str) -> str:
+    """Convert a reload cause abbreviation into its full descriptive string.
+
+    Examples
+    --------
+    ```python
+    >>> update_bgp_redistributed_proto_user("ZTP")
```
gmuloc, line 196:
```suggestion
    >>> convert_reload_cause("ZTP")
```

anta/custom_types.py
```diff
@@ -187,6 +187,23 @@ def update_bgp_redistributed_proto_user(value: str) -> str:
...
+    'System reloaded due to Zero Touch Provisioning'
+    ```
+    """
+    reload_causes = {"ztp": "System reloaded due to Zero Touch Provisioning"}
```
gmuloc, line 200: why is there only one entry in the dictionary? you should also have USER and FPGA

anta/tests/system.py
```diff
@@ -102,19 +93,9 @@ class VerifyReloadCause(AntaTest):
     class Input(AntaTest.Input):
         """Input model for the VerifyReloadCause test."""
 
-        allowed_causes: list[str] = Field(default=["USER", "FPGA"])
+        allowed_causes: list[ReloadCause] = Field(default=["Reload requested by the user.", "Reload requested after FPGA upgrade"])
```
gmuloc, line 96: I would expect this now
```suggestion
        allowed_causes: list[ReloadCause] = Field(default=["USER", "FPGA"])
```

tests/units/anta_tests/test_system.py
```diff
@@ -81,6 +81,25 @@
         "inputs": {"allowed_causes": ["ZTP"]},
         "expected": {"result": "success"},
     },
+    {
+        "name": "success-valid-reload-cause-fpga",
```
gmuloc, line 85: ADd a failure test where the inputs is 
```
"inputs": {"allowed_causes": ["ZTP"]},
```

and the eos data is FPGA or User to validate this scenario

tests/units/test_custom_types.py
```diff
@@ -267,3 +268,26 @@ def test_snmp_v3_prefix_valid_input() -> None:
     assert snmp_v3_prefix("auth") == "v3Auth"
     assert snmp_v3_prefix("noauth") == "v3NoAuth"
     assert snmp_v3_prefix("priv") == "v3Priv"
+
+
+@pytest.mark.parametrize(
+    ("str_input", "expected_output"),
+    [
+        pytest.param("ztp", "System reloaded due to Zero Touch Provisioning", id="valid"),
```
gmuloc, line 276: should also convert USER and FPGA

gmuloc (changes requested):

anta/tests/system.py
geetanjalimanegslab: Hi, @gmuloc 
If a default value is set, the custom type validator will not be triggered, and as a result, the value won't be converted into the expected description.

anta/custom_types.py
geetanjalimanegslab: If a default value is set, the custom type validator will not be triggered, and as a result, the value won't be converted into the expected description.

To avoid a breaking change, I have set the description as the default value for USER and FPGA.

anta/custom_types.py
geetanjalimanegslab: Updated Thanks.

anta/custom_types.py
geetanjalimanegslab: Thanks @gmuloc , Setting validate_default=True worked for me. I've applied the changes as per your suggestion.

tests/units/test_custom_types.py
geetanjalimanegslab: Add check for USER and FPGA

anta/custom_types.py
```diff
@@ -398,3 +415,7 @@ def snmp_v3_prefix(auth_type: Literal["auth", "priv", "noauth"]) -> str:
 NTPStratumLevel = Annotated[int, Field(ge=0, le=16)]
 PowerSupplyFanStatus = Literal["failed", "ok", "unknownHwStatus", "powerLoss", "unsupported"]
 PowerSupplyStatus = Literal["ok", "unknown", "powerLoss", "failed"]
+ReloadCause = Annotated[
+    str,
```
gmuloc, line 419:
```suggestion
    Literal["System reloaded due to Zero Touch Provisioning", "Reload requested by the user.", "Reload requested after FPGA upgrade"]
```

anta/tests/system.py
```diff
@@ -73,21 +73,29 @@ class VerifyReloadCause(AntaTest):
 
     Expected Results
     ----------------
-    * Success: The test will pass if there are NO reload causes or if the last reload was caused by the user or after an FPGA upgrade.
-    * Failure: The test will fail if the last reload was NOT caused by the user or after an FPGA upgrade.
+    * Success: The test passes if there are no reload causes, or if the last reload was user-initiated, after an FPGA upgrade, or caused by Zero Touch Provisioning.
+    * Failure: The test will fail if the last reload was NOT caused by the user or after an FPGA upgrade or due to Zero Touch Provisioning.
```
gmuloc, line 77:
```suggestion
    * Failure: The test will fail if the last reload cause was NOT one of the provided ones.
```

anta/tests/system.py
```diff
@@ -73,21 +73,29 @@ class VerifyReloadCause(AntaTest):
 
     Expected Results
     ----------------
-    * Success: The test will pass if there are NO reload causes or if the last reload was caused by the user or after an FPGA upgrade.
-    * Failure: The test will fail if the last reload was NOT caused by the user or after an FPGA upgrade.
+    * Success: The test passes if there are no reload causes, or if the last reload was user-initiated, after an FPGA upgrade, or caused by Zero Touch Provisioning.
```
gmuloc, line 76:
```suggestion
    * Success: The test passes if there are no reload cause, or if the last reload cause was one of the provided inputs.
```

gmuloc: Add the default allowed default causes in the docstring above

anta/tests/system.py
```diff
@@ -73,21 +73,29 @@ class VerifyReloadCause(AntaTest):
...
+    * Failure: The test will fail if the last reload was NOT caused by the user or after an FPGA upgrade or due to Zero Touch Provisioning.
     * Error: The test will report an error if the reload cause is NOT available.
 
     Examples
     --------
     ```yaml
     anta.tests.system:
       - VerifyReloadCause:
+        allowed_causes:
+          - ZTP
```
gmuloc, line 86:
```suggestion
          - USER
          - FPGA
          - ZTP
```

anta/custom_types.py
```diff
@@ -187,6 +187,23 @@ def update_bgp_redistributed_proto_user(value: str) -> str:
...
+    ```python
+    >>> convert_reload_cause("ZTP")
+    'System reloaded due to Zero Touch Provisioning'
+    ```
+    """
+    reload_causes = {"ztp": "System reloaded due to Zero Touch Provisioning", "user": "Reload requested by the user.", "fpga": "Reload requested after FPGA upgrade"}
+    if not reload_causes.get(value.lower()):
+        msg = f"Invalid reload cause: '{value}' - expected causes are {list(reload_causes)}"
+        raise ValueError(msg)
+    return reload_causes[value.lower()]
```
gmuloc, line 204:
```suggestion
    reload_causes = {"ZTP": "System reloaded due to Zero Touch Provisioning", "USER": "Reload requested by the user.", "FPGA": "Reload requested after FPGA upgrade"}
    if not reload_causes.get(value.upper()):
        msg = f"Invalid reload cause: '{value}' - expected causes are {list(reload_causes)}"
        raise ValueError(msg)
    return reload_causes[value.upper()]
```

tests/units/test_custom_types.py
```diff
@@ -267,3 +268,28 @@ def test_snmp_v3_prefix_valid_input() -> None:
...
+
+@pytest.mark.parametrize(
+    ("str_input"),
+    [
+        pytest.param("ztp2", id="invalid"),
+    ],
+)
+def test_invalid_convert_reload_cause(str_input: str) -> None:
+    """Test invalid convert_reload_cause."""
+    with pytest.raises(ValueError, match=r"Invalid reload cause: 'ztp2' - expected causes are \['ztp', 'user', 'fpga'\]"):
```
gmuloc, line 294:
```suggestion
    with pytest.raises(ValueError, match=r"Invalid reload cause: 'ztp2' - expected causes are \['ZTP', 'USER', 'FPGA'\]"):
```

gmuloc (changes requested):

gmuloc (approved):

carl-baillargeon (approved):
LGTM.

gmuloc (approved):

PR #1182, opened 2025-04-25
feat(anta.tests): Updated the VerifyVxlanVniBindings test to add support for VRF name.

### LINKED ISSUE #1164: feat(anta.tests): VerifyVxlanVniBindings should also support VRF name
L3VNIs use dynamic VLANs which can be hard to know in advance.

Test inputs can be updated to support the VRF name also:
```yaml
    anta.tests.vxlan:
      - VerifyVxlanVniBinding:
          bindings:
            10010: 10
            10020: 20
            500: PROD
            600: DEV
```
We can then check the `vniBindingsToVrf.<vni>.vrfName` key in the ouput.

geetanjalimanegslab:
# Description
Updated the VerifyVxlanVniBindings test to add support for VRF name.
Fixes #1164 

# Checklist:


vitthalmagadum (approved):

anta/tests/vxlan.py
```diff
@@ -127,7 +128,7 @@ class VerifyVxlanVniBinding(AntaTest):
     class Input(AntaTest.Input):
         """Input model for the VerifyVxlanVniBinding test."""
 
-        bindings: dict[Vni, VlanId]
+        bindings: dict[Vni, VlanId | str]
         """VNI to VLAN bindings to verify."""
```
carl-baillargeon, line 132:
```suggestion
        """VNI-VLAN or VNI-VRF bindings to verify."""
```

tests/units/anta_tests/test_vxlan.py
```diff
@@ -222,6 +222,24 @@
         "inputs": {"bindings": {10010: 10, 10020: 20, 500: 1199}},
         "expected": {"result": "failure", "messages": ["Interface: Vxlan1 VNI: 10010 - Binding not found"]},
     },
+    {
+        "name": "failure-vrf-wrong-binding",
```
carl-baillargeon, line 226: Can we also add a unit test if we have a VNI-VRF binding and the VNI is in the `vniBindings` output and not the `vniBindingsToVrf`? I think it would produce something like `Wrong VLAN binding - Expected: PROD Actual: 30` which is not what we want.

If a user gives a VNI-VRF binding as input, we should only check in the `vniBindingsToVrf` output.

carl-baillargeon (changes requested):

tests/units/anta_tests/test_vxlan.py
geetanjalimanegslab: Ohh yes! thanks @carl-baillargeon, Fixed this and added unit testcase for the same.

PR #1200, opened 2025-05-14
feat(anta.tests): Added test case to verify bad syslog events

vitthalmagadum:
# Description

Added test case to verify bad syslog events

Fixes #https://github.com/aristanetworks/avd-internal/issues/254

# Checklist:


anta/tests/logging.py
```diff
@@ -484,3 +499,44 @@ def test(self) -> None:
                 self.result.is_failure(
                     f"Pattern: `{logging_entry.regex_match}` - Not found in last {logging_entry.last_number_messages} {logging_entry.severity_level} log entries"
                 )
+
+
+class VerifyBadSyslog(AntaTest):
```
carl-baillargeon, line 504: Let's enhance the existing `VerifyLoggingEntries` test. In the `LoggingQuery` model, we can could do something like:
```
regex_match: RegexString | list[RegexString]
last_number_messages: Annotated[int, Field(ge=1, le=9999) | None = None
last_number_time_units: Annotated[int, Field(ge=1, le=9999) | None = None
time_unit: Literal["days", "hours", "minutes", "seconds"] = "days"
severity_level: LogSeverityLevel = "informational"
fail_on_match: bool = False
```
To follow SemVer, `last_number_messages` would take precedence over `last_number_time_units`. Depending on the inputs, you would render the appropriate `show logging ...` commands.

We can also add a validator to make sure one or the other is provided. We can also support a list of regex in the same query, so that the user can provide its own list of known bad keywords.

By default, the test should fail if the provided regex are not found (following current behavior). If `fail_on_match` is True, we actually fail if the regex are found.

anta/tests/logging.py
```diff
@@ -14,13 +14,28 @@
...
-from anta.custom_types import LogSeverityLevel
+from anta.custom_types import LogSeverityLevel, PositiveInteger
 from anta.input_models.logging import LoggingQuery
 from anta.models import AntaCommand, AntaTemplate, AntaTest
 
 if TYPE_CHECKING:
     import logging
 
+# Predefined error indicators
+KNOWN_BAD_KEYWORDS = [
```
carl-baillargeon, line 25: These should be user inputs (see my other long comment).

carl-baillargeon (changes requested):

anta/tests/logging.py
vitthalmagadum: Updated. Thanks!!

carl-baillargeon (approved):
LGTM if CI passes.

PR #1237, opened 2025-06-06
feat(anta): Add JUnit reporter option

### LINKED ISSUE #789: Add JUnit xml reports
JUnit XML report is a widely adopted format for many testing frameworks. The advantage over the currently available report formats in ANTA is the integration with tools such [gitlab](https://docs.gitlab.com/ee/ci/testing/unit_test_reports.html#view-unit-test-reports-on-gitlab) or [github](https://github.com/marketplace/actions/test-reporter) .

You can find more info about the format here: https://github.com/testmoapp/junitxml
And a python module to help with the generation: https://github.com/weiwei/junitparser

gmuloc:
# Description

Add JUnit output format when running NRFU

Fixes #789 

# TODOs:

- [ ] Need to test with same test run for the same device multiple time as right now the code does not handle this.

# Checklist:


carl-baillargeon (approved):
LGTM!

carl-baillargeon (changes requested):
Report doesn't work with duplicate tests (with different inputs).

gmuloc:
Closing this as the target platform (Gitlab) for this is not able to deduplicate tests based on key (at least this was not possible at the time of this PR writing) and so the results look correct but can be wrong, which is the worst case of wrong as it can go unnoticed and make users lose confidence in the report and so in the tool. Which we don't want for obvious reasons.

PR #1267, opened 2025-06-30
feat(anta.tests): Added test VerifyFlashUtilization to verify the free space percentage on the flash drive

geetanjalimanegslab:
# Description
Added test VerifyFlashUtilization to verify the free space percentage on the flash drive
Fixes # https://github.com/aristanetworks/avd-internal/issues/254

Added testcase - system_hardware_flash_free_space_test.py
# Checklist:

### DIFF
--- a/anta/tests/hardware.py
+++ b/anta/tests/hardware.py
@@ -7,14 +7,11 @@
 # mypy: disable-error-code=attr-defined
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, ClassVar, Literal
+from typing import Any, ClassVar, Literal
 
-from anta.custom_types import PositiveInteger, PowerSupplyFanStatus, PowerSupplyStatus
+from anta.custom_types import Percent, PositiveInteger, PowerSupplyFanStatus, PowerSupplyStatus
 from anta.decorators import skip_on_platforms
-from anta.models import AntaCommand, AntaTest
-
-if TYPE_CHECKING:
-    from anta.models import AntaTemplate
+from anta.models import AntaCommand, AntaTemplate, AntaTest
 
 
 class VerifyTransceiversManufacturers(AntaTest):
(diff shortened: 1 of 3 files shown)


anta/tests/hardware.py
```diff
@@ -379,3 +376,69 @@ def test(self) -> None:
...
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """Main test function for VerifyFlashUtilization."""
+        self.result.is_success()
+        command_outputs: dict[str, Any] = {}
+        command_outputs = {"primary supervisor": self.instance_commands[0].json_output}
+
+        # Collect the peer supervisor file-system details
+        if self.inputs.check_peer_supervisor:
+            command_outputs.update({"peer supervisor": self.instance_commands[1].json_output})
```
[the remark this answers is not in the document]
geetanjalimanegslab: hi @vitthalmagadum,  Incorporated all your suggestions in the PR please have a look.

anta/tests/hardware.py
```diff
@@ -379,3 +376,68 @@ def test(self) -> None:
...
+
+class VerifyFlashUtilization(AntaTest):
+    """Verifies the free space percentage on the flash drive.
+
+    Expected Results
+    ----------------
+    * Success: The test will pass if the flash utilization for the primary, and the peer supervisor (if check_peer_supervisor is True), is below the defined
+     threshold.
+    * Failure: The test will fail if the flash utilization for the primary, or for the peer supervisor (when check_peer_supervisor is True), is above the defined
+     threshold.
```
carl-baillargeon, line 389:
```suggestion
    """Verifies the free space on the flash drive is sufficient.
    
    !!! tip
        If `check_peer_supervisor` is True, the peer supervisor flash utilization is also verified.

    Expected Results
    ----------------
    * Success: The test will pass if flash utilization is below the defined threshold.
    * Failure: The test will fail if flash utilization is above the defined threshold.
```

anta/tests/hardware.py
```diff
@@ -379,3 +376,68 @@ def test(self) -> None:
...
+        """The maximum allowed percentage of flash memory utilization."""
+
+    def render(self, template: AntaTemplate) -> list[AntaCommand]:
+        """Render the template for peer supervisor."""
+        commands: list[AntaCommand] = []
+        if self.inputs.check_peer_supervisor:
+            commands.extend([template.render(check_peer_supervisor=""), template.render(check_peer_supervisor="session peer-supervisor ")])
+            return commands
+        commands.extend([template.render(check_peer_supervisor="")])
+        return commands
```
carl-baillargeon, line 421:
```suggestion
        """Render the template for the primary and optional peer supervisor."""
        if self.inputs.check_peer_supervisor:
            return [
                template.render(check_peer_supervisor=""),
                template.render(check_peer_supervisor="session peer-supervisor "),
            ]
        return [template.render(check_peer_supervisor="")]
```

anta/tests/hardware.py
```diff
@@ -379,3 +376,68 @@ def test(self) -> None:
...
+    anta.tests.hardware:
+      - VerifyFlashUtilization:
+          flash_utilization_threshold: 70
+          check_peer_supervisor: True # Optional
+    ```
+    """
+
+    categories: ClassVar[list[str]] = ["hardware"]
+    commands: ClassVar[list[AntaCommand | AntaTemplate]] = [
+        AntaTemplate(template="{check_peer_supervisor}show file systems", revision=1),
```
carl-baillargeon, line 403: Let's call the placeholder something else to avoid confusion with the actual input field name:
```suggestion
        AntaTemplate(template="{session_peer_supervisor}show file systems", revision=1),
```

anta/tests/hardware.py
```diff
@@ -379,3 +376,68 @@ def test(self) -> None:
...
+    ]
+
+    class Input(AntaTest.Input):
+        """Input model for the VerifyFlashUtilization test."""
+
+        check_peer_supervisor: bool = False
+        """If True, extends free space verification to the backup supervisor's flash drive."""
+        flash_utilization_threshold: Percent = 70
+        """The maximum allowed percentage of flash memory utilization."""
```
carl-baillargeon, line 412: I think we usually put boolean fields last:
```suggestion
        max_utilization: Percent = 70
        """The maximum allowed percentage of flash memory utilization."""
        check_peer_supervisor: bool = False
        """If True, also verifies the peer supervisor flash drive."""
```

anta/tests/hardware.py
```diff
@@ -379,3 +376,68 @@ def test(self) -> None:
...
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """Main test function for VerifyFlashUtilization."""
+        self.result.is_success()
+        command_outputs: dict[str, Any] = {}
+        command_outputs = {"primary": self.instance_commands[0].json_output}
+        # Collect the peer supervisor file-system details
+        if self.inputs.check_peer_supervisor:
+            command_outputs.update({"peer": self.instance_commands[1].json_output})
+        # TODO: Do we need to add validation of peer supervisor command output if check_peer_supervisor flag is True?
```
carl-baillargeon, line 433: Let's use a helper instead: `def _get_flash_utilization(self, file_system_details: dict[str, Any]) -> float | None`.

Here `file_system_details` would be the output of `show file systems` or `session peer-supervisor show file systems`.

You can then always check the primary flash utilization calling the helper, and then check the peer supervisor if requested (calling the helper again).

anta/tests/hardware.py
```diff
@@ -379,3 +376,68 @@ def test(self) -> None:
...
+        for supervisor, file_system_details in command_outputs.items():
+            for drive_details in file_system_details["fileSystems"]:
+                if drive_details["prefix"] == "flash:":
```
carl-baillargeon, line 436: Let's use `get_item` instead and have a failure if `flash:` dict is not found. This logic would be in the new helper.

anta/tests/hardware.py
carl-baillargeon, line 433: For the `session peer-supervisor` command:
If a device doesn't support a supervisor (regular switches), ANTA will catch the "not supported on this hardware platform".

If a device **does** support a supervisor but the command fails, we get this back from eAPI:
```
{
  "jsonrpc": "2.0",
  "id": "EapiExplorer-1",
  "error": {
    "code": 1000,
    "message": "CLI command 2 of 2 'session peer-supervisor show file systems' failed: could not run command",
    "data": [
      {},
      {
        "errors": [
          ""
        ]
      }
    ]
  }
}
```
Please test within ANTA (you can use DUT tg315) but I suspect the test would be error in that case.

anta/tests/hardware.py
```diff
@@ -379,3 +376,68 @@ def test(self) -> None:
...
+                    flash_utilization = round(100 - int(drive_details["free"]) / int(drive_details["size"]) * 100, 2)
+                    # Verify the free space percentage on the flash drive
+                    if flash_utilization > self.inputs.flash_utilization_threshold:
+                        self.result.is_failure(
+                            f"Supervisor: {supervisor.capitalize()} - Flash utilization is above threshold - "
+                            f"Expected: < {self.inputs.flash_utilization_threshold}% Actual: {flash_utilization}%"
```
carl-baillargeon, line 442: For dual-supervisor systems, it makes sense to have `Supervisor: Primary` and `Supervisor: Peer` as prefixes. Otherwise we shouldn't have any prefixes for regular systems:
```suggestion
                            f"Flash utilization above threshold - "
                            f"Expected: <= {self.inputs.flash_utilization_threshold}% Actual: {flash_utilization}%"
```
Separating the logic for regular/modular switches using the helper would solve this I think.

anta/tests/hardware.py
geetanjalimanegslab: As discussed, this will fix in `https://github.com/aristanetworks/anta/issues/1274
`

carl-baillargeon (approved):
LGTM if CI passes.

PR #1286, opened 2025-07-14
fix(anta.tests): Update recent tests failure messages with <= instead of <

### LINKED ISSUE #1278: fix(anta.tests). Update recent tests failure messages with <= instead of <

vitthalmagadum:
# Description

fix(anta.tests). Update recent tests failure messages with <= instead of <

Fixes #1278


carl-baillargeon (changes requested):
I think `VerifyInterfacesVoqAndEgressQueueDrops` is missing. Please double-check we are not missing anything :)

PR #1306, opened 2025-08-14
feat(anta): Added pre commit hook to validate examples/tests.yaml

vitthalmagadum:
# Description

Added pre commit hook to validate examples/tests.yaml


gmuloc:
why do you need this? the file is auto generted by pre-commit - if it fails it means the wrong ilfe is generated. Can you please explain more?

PR #1463, opened 2026-03-18
feat(anta.tests): add VerifySSHAlgorithms test

brady-anet:
## Summary

  - Adds `VerifySSHMACAlgorithms` to `anta/tests/security.py` to verify that management SSH MAC
    algorithms are explicitly configured as `hmac-sha2-256` and `hmac-sha2-512` only, using
    `show running-config | section management ssh`
  - Adds three unit test cases: success, wrong algorithms configured, and no MAC line present

  ## Why a new test rather than using an existing one?

  **`VerifySSHFIPSRestrictions` does not cover this.**
  `VerifySSHFIPSRestrictions` checks the runtime FIPS mode status from `show management ssh`
  (i.e., `FIPS status: enabled/disabled`). It does not inspect which MAC algorithms are explicitly
  configured in the running-config. A device with FIPS mode disabled could still have the correct
  MAC algorithms configured — and vice versa — so these are distinct checks.

  **`VerifyRunningConfigLines` is not appropriate.**
  `VerifyRunningConfigLines` fetches and regex-searches the entire running-config. The test itself
  carries a built-in warning to avoid it when a more specific test exists. Additionally, it can only
  assert presence of a pattern — it cannot enforce that no other (unapproved) MAC algorithms are
  present alongside the required ones. `VerifySSHMACAlgorithms` uses set equality, so a superset
  like `hmac-sha1 hmac-sha2-256 hmac-sha2-512` correctly fails.

  **`VerifySSHFIPSRestrictions` and `VerifySSHAlgorithms` should not be merged into one.**
  They use different EOS commands (`show management ssh` vs `show running-config | section management ssh`)
  because they check different properties: runtime daemon state vs persisted configuration. Combining
  them would conflate two separate controls, reduce failure message clarity, and require two commands
  in a single test.

**`VerifySSHAlgorithms` has been generalized to support both mac (V-255960) and cipher (V-255961) use cases.

  ## Checklist

  - [x] I have run pre-commit for code linting and typing (`pre-commit run`) — ruff and pylint clean, `examples/tests.yaml` auto-updated
  - [x] I have made corresponding changes to the documentation — docs are auto-generated from the class docstring via mkdocstrings; `Expected Results` and `Examples` sections are present
  - [x] I have added tests that prove my fix is effective or that my feature works — 3 unit test cases added (success, wrong algorithms, no MAC line)
  - [x] New and existing unit tests pass locally with my changes (`tox -e py312`) — 1968 passed, 1 skipped

  🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,8 @@ venv.bak/
 
 # VScode settings
 .vscode
+
+# Local development files
+CLAUDE.md
+inventory.yml
+.idea/
--- a/anta/tests/security.py
+++ b/anta/tests/security.py
@@ -776,6 +776,48 @@ class VerifySSHFIPSRestrictions(AntaTest):
             self.result.is_success()
 
 
+class VerifySSHMACAlgorithms(AntaTest):
+    """Verifies that management SSH MAC algorithms are configured to only FIPS-approved HMAC algorithms.
+
+    Checks the running configuration to confirm that the management SSH MAC algorithms are
+    explicitly set to hmac-sha2-256 and hmac-sha2-512, as required by STIG V-255960.
+
+    Expected Results
+    ----------------
+    * Success: The test will pass if MAC algorithms are configured as exactly hmac-sha2-256 and hmac-sha2-512.
+    * Failure: The test will fail if the MAC algorithms are not configured or include unapproved algorithms.
+
+    Examples
+    --------
+    ```yaml
+    anta.tests.security:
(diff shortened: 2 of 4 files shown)


carl-baillargeon (changes requested):
Blocking this until we find a solution to parse the running-config.

carl-baillargeon:
@brady-anet You can now use the new `VerifyRunningConfig` test.

PR #1492, opened 2026-05-25
feat: Add session support to asynceapi

### LINKED ISSUE #1485: feat: Add session support to asynceapi
eAPI Cookies documentation is available in eAPI Explorer of an EOS device: `https://<device_ip>/eapi/cookies`.

High level implementation plan:

1. Add `use_session: bool = False` kwarg to `asynceapi.Device`. Store as private attribute.

2. Add `async def login(self) -> None`:  
  - POSTs `{"username": ..., "password": ...}` to `/login`
  - On 200: cookie should be captured automatically by HTTPX cookie jar so need to clear `self.auth` so Basic auth headers stop being sent alongside the cookie
  - On 401: raise a new `EapiAuthenticationError`
  - If already logged in (cookie present and new flag `_logged_in=True`), nothing to do
  - Probably need to be guarded by an internal Lock to avoid race condition between concurrent first requests

3. Add `async def logout(self) -> None`:
  - POSTs to `/logout`. Maybe log non-2XX, TBD
  - Clears `_logged_in`

4. Update `jsonrpc_exec` and `check_api_endpoint` to call `await self.login()` first if using a session and not logged in. ~Maybe retry once on a 401 response. If second attempt also 401 then we raise.~ _EDIT: retry will be considered in a future PR_
 
5. Override `aclose()`:
  - If using session and logged in, call `await self.logout()`
  - Close HTTPX with super
  
6. Username/password must be stored as instance attributes since they will be used in `login()` payload. (Please consider the security aspect of this and make sure passwords are not leaked). Maybe we can avoid storing them as instance attributes and extract them from BasicAuth instead... Worth exploring: https://www.python-httpx.org/advanced/authentication/#custom-authentication-schemes

For tests:
- Login success
- Login 401
- Command after login uses cookie
- ~401 then relogin retry~
- aclose() calls logout

geetanjalimanegslab:
# Description
Added support for asynceapi, using use_session knob.

Fixes #1485

# Checklist:


asynceapi/device.py
```diff
@@ -476,6 +500,53 @@ async def jsonrpc_exec(self, jsonrpc: JsonRpc) -> list[EapiJsonOutput] | list[Ea
...
+        try:
+            response = await self.post(self.EAPI_LOGOUT_URL, auth=httpx.Auth(), headers=self._session_auth.cookie_header)
+            if not response.is_success:
+                LOGGER.warning("Logout returned non-2xx status %s for %s", response.status_code, self.host)
+        except httpx.HTTPError as exc:
+            LOGGER.warning("Logout HTTP error for %s: %s", self.host, exc)
+        finally:
+            self._session_auth.reset()
+
+    async def __aexit__(
```
gmuloc, line 523: `Device.__aexit__()` logs out for `async with`, but `aclose()` is still inherited from
  `httpx.AsyncClient` and does not call `/logout`. A caller using the standard HTTPX
  manual lifecycle:

```python
device = Device(...)
try:
finally: 
  await device.aclose()
```

  will close the transport but leave the EOS cookie session alive server-side until
  the device session timeout expires.

Can we override `aclose()` to perform the same
  logout-before-close workflow, and have `__aexit__()` reuse that path?


```python
  class Device(httpx.AsyncClient):
      async def aclose(self) -> None:
          if self._use_session:
              await self.logout()
          await super().aclose()

      async def __aexit__(self, exc_type=None, exc_value=None, traceback=None) -> None:
          await self.aclose()
```


right now we handle well only the context manager case

asynceapi/_auth.py
```diff
@@ -0,0 +1,263 @@
...
+        Raises
+        ------
+        EapiAuthenticationError
+            If the response status code is 401 Unauthorized.
+        httpx.HTTPStatusError
+            If the response status code is any other non-2xx value.
+        """
+        validate_login_status(response.status_code, self._host)
+        response.raise_for_status()
+        self._state.session_cookie = get_session_cookie(response.cookies)
```
gmuloc: if this returns None, we should raise as we will show ourselves as logged in without a cookie and things will be sad

gmuloc (changes requested):

asynceapi/_auth.py
gmuloc: we could do a retry maybe but otherwise I feel like raising is the most appropriate

asynceapi/errors.py
```diff
@@ -49,3 +49,11 @@ def __str__(self) -> str:
 
 # alias for exception during sending-receiving
 EapiTransportError = httpx.HTTPStatusError
+
+
+class EapiAuthenticationError(RuntimeError):
+    """Exception raised when eAPI session login returns 401."""
+
+    def __init__(self, host: str) -> None:
+        super().__init__(f"Authentication failed for {host!r} (HTTP 401): verify credentials and that session-based authentication is enabled on the device")
```
gmuloc:
```suggestion
        super().__init__(f"Authentication failed for {host!r} (HTTP 401): verify your credentials")
```

gmuloc: @geetanjalimanegslab please verify but there is nothing to enable on the device

asynceapi/device.py
geetanjalimanegslab: Thanks Guillaume for the suggestion, Updated.

gmuloc (commented):
to make pre-commit pass add `asend` to https://github.com/aristanetworks/anta/blob/main/.codespellignore

tests/units/asynceapi/test__auth.py
```diff
@@ -0,0 +1,342 @@
+# Copyright (c) 2024-2026 Arista Networks, Inc.
```
gmuloc:
```suggestion
# Copyright (c) 2026 Arista Networks, Inc.
```

tests/units/asynceapi/test__auth.py
```diff
@@ -0,0 +1,342 @@
...
+    """Return a fresh EapiSessionAuth with known credentials."""
+    return EapiSessionAuth(username=_USERNAME, password=_PASSWORD, login_url=_LOGIN_URL, host=_HOST)
+
+
+# ---------------------------------------------------------------------------
+# Pure functions
+# ---------------------------------------------------------------------------
+
+
+class TestBuildLoginPayload:  # pylint: disable=too-few-public-methods
```
gmuloc: please don't do classes for these test s- we don't do it in another places. (and you can remove all the pylint stuff)

asynceapi/_auth.py
```diff
@@ -0,0 +1,269 @@
...
+    def reset(self) -> None:
+        """Set ``logged_in`` to False and clear ``session_cookie``."""
+        self.logged_in = False
+        self.session_cookie = None
+
+
+# Library-agnostic utilities (no httpx dependency)
+
+
+def build_login_payload(username: str, password: str) -> dict[str, str]:
```
gmuloc: for these functiopns which are oneliners (with 10 lines of docstrings), lets inline them in the code. This is too much, too wordy to maintain. Please update the tests as well.

As a rule of thumb a helper function should be created only when called multiple times

asynceapi/_auth.py
```diff
@@ -0,0 +1,269 @@
...
+    ----------
+    cookies
+        The response cookie mapping (e.g. ``response.cookies``).
+
+    Returns
+    -------
+    str | None
+        The Session cookie value, or None if the cookie is absent.
+    """
+    return cookies.get("Session")
```
gmuloc: this oneliner should be gone

asynceapi/_auth.py
```diff
@@ -0,0 +1,269 @@
...
+    session_cookie
+        The session cookie value returned by EOS after a successful login,
+        or None if no session is active.
+
+    Returns
+    -------
+    dict[str, str]
+        ``{"Cookie": "Session=<value>"}`` when a cookie is present, empty dict otherwise.
+    """
+    return {"Cookie": f"Session={session_cookie}"} if session_cookie else {}
```
gmuloc: lets build the header where we need it

asynceapi/_auth.py
```diff
@@ -0,0 +1,263 @@
...
+        response.raise_for_status()
+        self._state.session_cookie = get_session_cookie(response.cookies)
+        self._state.logged_in = True
+
+    def _attach_cookie(self, request: httpx.Request) -> None:
+        """Attach the active session cookie to an outgoing request header in place."""
+        if self._state.session_cookie:
+            request.headers["Cookie"] = f"Session={self._state.session_cookie}"
+
+    async def async_auth_flow(
```
gmuloc, line 54: according to doc you should also override sync_auth_flow and raise:

https://www.python-httpx.org/advanced/authentication/#custom-authentication-schemes (cf art the end)

asynceapi/_auth.py
```diff
@@ -0,0 +1,263 @@
...
+                    LOGGER.debug("[AUTH] login OK — cookie=%s...", (self._state.session_cookie or "")[:8])
+
+        # Phase 2: attach session cookie and dispatch the real request
+        self._attach_cookie(request)
+        response = yield request
+
+        # Phase 3: mid-session 401 — re-login once and retry
+        if response.status_code == HTTPStatus.UNAUTHORIZED:
+            LOGGER.debug("[AUTH] got 401 — resetting session, will re-login")
+            self.reset()
```
gmuloc: reseting outside of the lock could cause race conditions if another coroutine just updated the cookie

asynceapi/_auth.py
```diff
@@ -0,0 +1,269 @@
...
+                if not self._state.logged_in:
+                    login_response = yield self._build_login_request()
+                    self._handle_login_response(login_response)
+                    LOGGER.debug("[AUTH] login OK — cookie=%s...", (self._state.session_cookie or "")[:8])
+
+        # Phase 2: attach session cookie and dispatch the real request
+        self._attach_cookie(request)
+        response = yield request
+
+        # Phase 3: mid-session 401 — re-login once and retry
```
gmuloc: after discussion with @carl-baillargeon, this is opening a lot of potential issues - lets ignore for now the retry and just fail is the cookie has expired.

lets do it in two phases

1/ no retry - we jsust fail if the session cookie has expired (we can give a meanmingful tentative error message for this)

2/ we will look at retrying but there are more state we need to catch with the current implementation as we could end up with re-login storms.

gmuloc: at least two race conditions possible with current code:

1/ reset() in one coroutine erase a fresh cookie in another corouting
2/ checking _state.logged_in outside of the lock can lead to seeing it as True in a coroutine while another that just experience a 401 is calling reset.

so for now lets be a bit simpler and avoid these corner cases

gmuloc (changes requested):

asynceapi/_auth.py
geetanjalimanegslab: Yes  agreed, removed. Thanks

asynceapi/_auth.py
geetanjalimanegslab: sure updated.

asynceapi/_auth.py
geetanjalimanegslab: sure updated.

asynceapi/_auth.py
geetanjalimanegslab: Great catch! This is handled in the current changes — sync_auth_flow is overridden and raises EapiAsyncOnlyError to explicitly block usage with synchronous clients.

asynceapi/_auth.py
geetanjalimanegslab: Thanks @gmuloc, Agreed on the two-phase approach. In Phase 1, both race conditions are avoided — a 401 immediately raises EapiAuthenticationError with no reset() triggered, and reset() itself acquires the same asyncio.Lock as login, so a concurrent reset can never erase a cookie mid-login. Would love to hear your thoughts on this approach!

asynceapi/_auth.py
geetanjalimanegslab: Yes, understood. I have updated the PR with the required changes. Thanks.

asynceapi/_auth.py
geetanjalimanegslab: Yes agreed, updated Thanks.

asynceapi/errors.py
geetanjalimanegslab: You are absolutely correct! there is nothing to enable  on the device. updated error message.

Thanks.

tests/units/asynceapi/test__auth.py
geetanjalimanegslab: Yes agreed and removed test classes.

tests/integration/test_device_session.py
```diff
@@ -0,0 +1,72 @@
...
+}
+
+_COMMAND_RESPONSE = {
+    "jsonrpc": "2.0",
+    "id": "EapiExplorer-1",
+    "result": [{"modelName": "pytest"}],
+}
+
+
+async def test_device_session_logout_sends_cookie_and_resets_state() -> None:
```
gmuloc: let's move these to units/asynceapi/test_device.py

gmuloc:
Let's cover the missing lines in device but not in _auth (these are non obvious to create conditions)

gmuloc (commented):
@geetanjalimanegslab this looks good to me so far with what we have discussed. The way forward will be for ANTA to handle calling the login again on 401 expiry so we can decide there using a lock who is doing what. Please add the tests we discussed yesterday

asynceapi/_auth.py
gmuloc: the good solution will then be to handle all the recovery stuff at a higher level - probably at AntaDevice level - we will see

asynceapi/device.py
```diff
@@ -501,57 +499,27 @@ async def jsonrpc_exec(self, jsonrpc: JsonRpc) -> list[EapiJsonOutput] | list[Ea
...
-        Does nothing if no session is active (``use_session=False`` or not yet logged in).
-        Session state is always cleared regardless of the server response, so the client
-        is never left in a half-logged-in state. HTTP and transport errors are logged as
-        warnings and not re-raised.
-        """
-        LOGGER.debug("[AUTH] logout() called — logged_in=%s", self._session_auth.logged_in if self._session_auth else "N/A")
+        """Log out of the device session and reset local state. No-op if not logged in."""
         if self._session_auth is None or not self._session_auth.logged_in:
             return
+        cookie = self._session_auth.session_cookie  # capture before await — safe against concurrent reset()
```
gmuloc: maybe lets remove the comment here

asynceapi/_auth.py
```diff
@@ -0,0 +1,88 @@
...
+from typing import TYPE_CHECKING
+
+import httpx
+
+from .errors import EapiAsyncOnlyError, EapiAuthenticationError
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncGenerator, Generator
+
+LOGGER = getLogger(__name__)
```
carl-baillargeon: Not used in this module.

asynceapi/device.py
```diff
@@ -99,11 +109,23 @@ def __init__(
...
-
-        auth_object = httpx.BasicAuth(username, password) if username and password else None
-        kwargs.setdefault("auth", auth_object)
+        if self._use_session:
+            if not (username and password):
+                msg = "username and password are required for session authentication"
+                raise ValueError(msg)
+            if not self.host:
+                msg = "host is required for session authentication"
+                raise ValueError(msg)
```
carl-baillargeon, line 124: We could simplify this and have a single raise.

asynceapi/__init__.py
```diff
@@ -5,8 +5,9 @@
 
 """Arista EOS eAPI asyncio client."""
 
+from ._auth import EapiSessionAuth
```
carl-baillargeon: I don't think we should expose this class, in case we need to change something down the road.

@gmuloc

asynceapi/__init__.py
```diff
@@ -5,8 +5,9 @@
...
 from .config_session import SessionConfig
 from .device import Device
-from .errors import EapiCommandError
+from .errors import EapiAuthenticationError, EapiCommandError
```
carl-baillargeon: Same thing here. Let's wait until we do the implementation on the ANTA side.

asynceapi/device.py
```diff
@@ -62,6 +67,8 @@ def __init__(
         password: str | None = None,
         proto: str = "https",
         port: str | int | None = None,
+        *,
+        use_session: bool = False,
```
carl-baillargeon, line 73: Should we put this in preview somehow while we complete the implementation in ANTA?

@gmuloc

asynceapi/__init__.py
gmuloc: agreed let's not expose for now. (I don't think it really matters anyways but at least it is hidden in _auth)

gmuloc:
```suggestion
```

asynceapi/__init__.py
gmuloc:
```suggestion
from .errors import EapiCommandError
```

asynceapi/__init__.py
```diff
@@ -5,8 +5,9 @@
...
 
-__all__ = ["Device", "EapiCommandError", "SessionConfig"]
+__all__ = ["Device", "EapiAuthenticationError", "EapiCommandError", "EapiSessionAuth", "SessionConfig"]
```
gmuloc:
```suggestion
__all__ = ["Device", "EapiCommandError", "SessionConfig"]
```

asynceapi/_auth.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) 2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+"""asynceapi session-based (cookie) authentication."""
+
+from __future__ import annotations
+
+import asyncio
+from http import HTTPStatus
+from logging import getLogger
```
gmuloc:
```suggestion
```

asynceapi/_auth.py
gmuloc:
```suggestion
```

asynceapi/device.py
gmuloc: nah - people are not supposed to consume this unexisting package :)

asynceapi/device.py
gmuloc: well I almost commented the opposite to have more specific error messages and also right now if username or password is `0` (I KNOW UNLIKELY) the code will fail

carl-baillargeon: I don't really care actually :P

gmuloc: lets leave it as it is

asynceapi/device.py
gmuloc:
```suggestion
        cookie = self._session_auth.session_cookie
```

asynceapi/_auth.py
```diff
@@ -0,0 +1,85 @@
...
+    def __repr__(self) -> str:
+        return f"EapiSessionAuth(host={self._host!r}, logged_in={self.logged_in})"
+
+    async def reset(self) -> None:
+        """Invalidate the current session, waiting for any in-progress login to complete first."""
+        async with self._lock:
+            self.session_cookie = None
+
+    def sync_auth_flow(self, request: httpx.Request) -> Generator[httpx.Request, httpx.Response, None]:  # noqa: ARG002
+        """Not supported — this auth class requires an async httpx client."""
```
carl-baillargeon:
```suggestion
    def sync_auth_flow(self, request: httpx.Request) -> Generator[httpx.Request, httpx.Response, None]:
        """Not supported — this auth class requires an async httpx client."""
        _ = request
```

asynceapi/_auth.py
```diff
@@ -0,0 +1,85 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncGenerator, Generator
+
+
+class EapiSessionAuth(httpx.Auth):
+    """httpx.Auth implementation for eAPI cookie-session authentication.
+
+    Performs a single login on the first request and attaches the session cookie thereafter.
+    A 401 on any request raises EapiAuthenticationError immediately. Re-login is not supported.
```
carl-baillargeon: After an explicit `logout()` -> `reset()`, a new command successfully re-logs in with a fresh cookie.
```suggestion
    A 401 on any request raises EapiAuthenticationError immediately.
```

asynceapi/device.py
```diff
@@ -476,6 +498,30 @@ async def jsonrpc_exec(self, jsonrpc: JsonRpc) -> list[EapiJsonOutput] | list[Ea
...
+
+    async def aclose(self) -> None:
+        """Log out and close the underlying HTTPX transport."""
+        if self._use_session:
+            await self.logout()
+        await super().aclose()
+
+    async def __aexit__(self, *_: object) -> None:
+        """Log out and close on context-manager exit."""
+        await self.aclose()
```
carl-baillargeon: We are losing exception context with this override. The parent `httpx.AsyncClient.__aexit__` passes `(exc_type, exc_value, traceback)` to the transport `__aexit__`. This override calls `self.aclose()` -> `super().aclose()`, which uses `transport.aclose()` instead so the exception info is not propagated.
```suggestion
    async def __aexit__(self, *args: object) -> None:
        """Log out and close on context-manager exit."""
        if self._use_session:
            await self.logout()
        await super().__aexit__(*args)
```

carl-baillargeon: Maybe we can improve the signature too with exc_type, exc_value, traceback.

asynceapi/device.py
```diff
@@ -83,6 +90,9 @@ def __init__(
             If not provided, the proto value is used to look up the associated
                   port (http=80, https=443). If provided, overrides the port used to
                   communite with the device.
+        use_session
```
carl-baillargeon, line 95: Typo above. "communite" -> "communicate".

carl-baillargeon (approved):
LGTM! Tested against a real device via a script. Results:
```
Target: 10.18.189.134 (user=cvpadmin)
============================================================

=== Test 1: Basic Auth (baseline) ===
  PASS: model=vEOS-lab version=4.34.3M

=== Test 2: Session Auth login + command ===
  PASS: cookie='fadc384a-fc85-451a-923c-c79a4023a55e' model=vEOS-lab

=== Test 3: Context manager lifecycle ===
  Inside context: hostname=DC1-BL1, logged_in=True
  PASS: After __aexit__: logged_in=False

=== Test 4: Concurrent commands (single login) ===
  PASS: Got 5 results with a single session cookie

=== Test 5: Bad credentials ===
  PASS: Got expected: Authentication failed for '10.18.189.134' (HTTP 401).

=== Test 7: Explicit logout + re-use ===
  Before logout: logged_in=True
  After logout: logged_in=False
  PASS: Re-login after explicit logout succeeded, new cookie='1ba4dd0a-2622-4688-9614-6b78e494003d'

--- Starting timeout test (will wait ~75s) ---
    Press Ctrl+C to skip

=== Test 6: Session timeout (requires 1-minute eAPI timeout) ===
  First command OK (cookie='2010ab84-536e-4f14-8eab-9ee4d9a313b5')
  Waiting 75s for session to expire on the device...
  PASS: Got expected timeout error: Authentication failed for '10.18.189.134' (HTTP 401).
Logout returned non-2xx status 400 for 10.18.189.134

============================================================
Done.
```

gmuloc (approved):

PR #1499, opened 2026-06-08
feat(anta.tests): Update VerifyTransceiversManufacturers

### LINKED ISSUE #1495: feat(anta.tests): Update VerifyTransceiversManufacturers
VerifyTransceiversManufacturers should have a boolean input that allows to ignore non present. Today we recommend using `"Not Present"` as a manufacturer

```diff
     anta.tests.hardware:
       - VerifyTransceiversManufacturers:
           manufacturers:
-            - Not Present           # this will still work
             - Arista Networks
             - Arastra, Inc.
+          allow_not_present: <bool, default=False> # False to be non breaking    
```

If `allow_not_present is true`, transceiver where manufacturer is "Not Present" should be skipped (`continue`)

geetanjalimanegslab:
## Description
Updated VerifyTransceiversManufacturers with adding **allow_not_present**  parameter if it. is true, transceiver where manufacturer is "Not Present" should be skipped (continue)

Fixes #1495 

## Checklist

### DIFF
--- a/anta/tests/hardware.py
+++ b/anta/tests/hardware.py
@@ -54,7 +54,7 @@ class VerifyTransceiversManufacturers(AntaTest):
         manufacturers: list[str]
         """List of approved transceivers manufacturers."""
         allow_not_present: bool = False
-        """If True, transceivers with manufacturer `Not Present` are skipped. Defaults to False."""
+        """Allow ports with no transceiver present"""
 
     @skip_on_platforms(["cEOSLab", "vEOS-lab", "cEOSCloudLab", "vEOS"])
     @AntaTest.anta_test
--- a/anta/tests/hardware.py
+++ b/anta/tests/hardware.py
@@ -38,9 +38,9 @@ class VerifyTransceiversManufacturers(AntaTest):
     anta.tests.hardware:
       - VerifyTransceiversManufacturers:
           manufacturers:
-            - Not Present
             - Arista Networks
             - Arastra, Inc.
+          allow_not_present: true
     ```
     """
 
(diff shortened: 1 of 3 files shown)


anta/tests/hardware.py
```diff
@@ -62,15 +64,18 @@ def test(self) -> None:
         command_output = self.instance_commands[0].json_output
 
         for port, value in command_output["xcvrSlots"].items():
+            mfg_name = value["mfgName"]
+
+            if self.inputs.allow_not_present and mfg_name == "Not Present":
+                continue
```
vitthalmagadum, line 73: I think this should be after atomic result. So we can have results for this.

anta/tests/hardware.py
```diff
@@ -53,6 +53,8 @@ class Input(AntaTest.Input):
 
         manufacturers: list[str]
         """List of approved transceivers manufacturers."""
+        allow_not_present: bool = False
+        """If True, transceivers with manufacturer `Not Present` are skipped. Defaults to False."""
```
carl-baillargeon:
```suggestion
        """Allow ports with no transceiver present."""
```

anta/tests/hardware.py
carl-baillargeon: I don't think we have proper support for skipped atomic results, but still I don't think we should add them. It doesn't bring more value to the users IMO.

anta/tests/hardware.py
geetanjalimanegslab: Updated Thanks.

anta/tests/hardware.py
geetanjalimanegslab: Ok. so skipping this change for now.

vitthalmagadum (approved):

anta/tests/hardware.py
```diff
@@ -53,6 +53,8 @@ class Input(AntaTest.Input):
 
         manufacturers: list[str]
         """List of approved transceivers manufacturers."""
+        allow_not_present: bool = False
+        """Allow ports with no transceiver present"""
```
carl-baillargeon:
```suggestion
        """Allow ports with no transceiver present."""
```

carl-baillargeon (approved):

PR #1509, opened 2026-06-15
feat(anta): Add support for session to AsyncEOSDevice

vitthalmagadum:
## Description

- Added use_session: bool = False to ANTA inventory models for hosts, networks, and ranges.
- anta/cli/utils.py — CLI flag --use-session-auth / --no-session-auth (tri-state: True | False | None)
- New AntaDeviceCapabilities dataclass
- EapiAuthenticationError now handled in collect()
- log level fix

Fixes  #1508 

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have run pre-commit for code linting and typing (`pre-commit run`)
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes (`tox -e testenv`)

## Summary by CodeRabbit

* **New Features**
  * Added optional session-based eAPI authentication for device connections.
  * Inventory entries can enable it per host, network, or IP range.
  * Added a global `--use-session` CLI option (and `ANTA_USE_SESSION`) to enable it across runs, overriding per-device inventory values.
* **Bug Fixes**
  * Ensured session mode is reflected when exporting/dumping inventories.
  * When disconnecting, session-enabled devices now properly log out before closing.
* **Documentation**
  * Updated developer and CLI help docs to describe the new option.
* **Tests**
  * Expanded unit tests to cover precedence, parsing, dumping, and logout behavior.

PR #1517, opened 2026-06-22
fix(anta): ResourceWarning when ANTA CLI is completed

vitthalmagadum:
## Description

Renamed `_session to _client`
Extracted the client creation from __init__ to a private method.
If the httpx client (self._client) is closed then restarted in refresh.
In _collect, if the client is closed, raised and direct people to call refresh first.
Implemented a disconnect function on the AsyncEOSDevice to close the _client.

Fixes #1442

## Checklist

## Release Notes

* **New Features**
  * Added a `--disconnect` option for `anta nrfu` (supports `ANTA_DISCONNECT_INVENTORY`) to disconnect devices after the run.
  * Added device-level `disconnect()` and inventory-wide `disconnect_inventory()` support.

* **Improvements**
  * Enhanced EOS eAPI connection lifecycle with lazy client creation and automatic recreation after refresh.
  * Updated tech-support command execution for the `--configure` path.

* **Bug Fixes / Tests**
  * Updated unit tests to reflect the new disconnect/refresh behavior and closed-client handling.
  * Adjusted settings propagation tests for `ANTA_HTTPX_TRUST_ENV`.

* **Documentation**
  * Updated `anta nrfu --help` snippet to include `--disconnect`.


geetanjalimanegslab (approved):
LGTM!

vitthalmagadum:
@coderabbitai review

anta/device.py
```diff
@@ -592,9 +634,21 @@ async def refresh(self) -> None:
...
 
+    async def disconnect(self) -> None:
+        """Close the eAPI httpx client.
+
+        Safe to call even if the client is already closed.
+        Use `refresh()` to reconnect.
+        """
+        logger.debug("Disconnecting device %s", self.name)
+        if not self._client.is_closed:
+            await self._client.aclose()
```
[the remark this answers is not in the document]
vitthalmagadum: The invocation of disconnect() will be added in a separate PR.
Thanks!

anta/device.py
```diff
@@ -327,6 +330,14 @@ class AsyncEOSDevice(AntaDevice):
...
     tags : set[str]
         Tags for this device.
+    enable : bool
+        When True, commands are collected in privileged (enable) mode.
+    _client : asynceapi.Device
+        The underlying HTTPX-based eAPI client. Recreated by `_create_client()`.
+        Closed by `disconnect()`; automatically recreated on the next `refresh()` call.
+    _ssh_opts : SSHClientConnectionOptions
+        SSH connection configuration (host, port, credentials, host-key policy).
+        Used to establish transient SSH connections in `copy()`.
```
gmuloc: is thissuggested by AI?

gmuloc: there are many other private attributes so maybe lets not advertise these as class public API. (_host, _username, ...)

anta/device.py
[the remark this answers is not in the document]
gmuloc: @vitthalmagadum this PR is to suppress the ResourceWarning so you have to call disconnect

anta/device.py
```diff
@@ -592,9 +634,21 @@ async def refresh(self) -> None:
...
             self.established = True
 
+    async def disconnect(self) -> None:
+        """Close the eAPI httpx client.
+
+        Safe to call even if the client is already closed.
+        Use `refresh()` to reconnect.
+        """
+        logger.debug("Disconnecting device %s", self.name)
+        if not self._client.is_closed:
```
gmuloc, line 668: Should disconnect also impact the os_online, established ... )at least established no? we are not established anymore when the client is disconnected)

anta/device.py
```diff
@@ -394,9 +405,13 @@ def __init__(  # noqa: PLR0913
             raise ValueError(message)
         self.enable = enable
         self._enable_password = enable_password
-        self._session: asynceapi.Device = asynceapi.Device(
-            host=host, port=port, username=username, password=password, proto=proto, timeout=timeout, trust_env=get_httpx_settings().trust_env
-        )
+        self._host = host
```
gmuloc: hmmm do we have a better way of storing this maybe? rather than many private attributes on the class. Maybe a little frozen dataclass or something else?

gmuloc (changes requested):

anta/device.py
vitthalmagadum: Yes, I use Claude to rephrase my statements, though I have been checking whether it is appropriate to include them here given that they are private attributes. Should we keep these (_client, _ssh_opts) as they are? In the meantime, I will look into using a dataclass for them.

anta/device.py
vitthalmagadum: Calling disconnect won't affect those, but I think we should mark them as offline there as well. I'll check how the benchmark behaves afterward.

anta/device.py
vitthalmagadum: Added dataclass. Thanks!!

anta/device.py
vitthalmagadum: Should we mark them as offline?
@gmuloc @carl-baillargeon

anta/device.py
```diff
@@ -292,6 +293,14 @@ async def refresh(self) -> None:
         - `hw_model`: The hardware model of the device.
         """
 
+    @abstractmethod
```
gmuloc: we cannot add a new abstract method as this would be breaking - we should instead raise a NotImplementedError if this is being called (and in our runner, we should catch NotImplementedError and print a helpful log and ignore the error for now)

anta/inventory/__init__.py
```diff
@@ -378,6 +378,17 @@ def is_base_class(self, device: AntaDevice) -> TypeIs[AntaDevice]:
...
         return not hasattr(device, "host") and not hasattr(device, "port")
 
+    async def disconnect_inventory(self) -> None:
+        """Run `disconnect()` coroutines for all AntaDevice objects in this inventory."""
+        results = await asyncio.gather(
+            *(device.disconnect() for device in self.values()),
+            return_exceptions=True,
+        )
+        for r in results:
+            if isinstance(r, Exception):
```
gmuloc, line 388: handle NotImplementedError here to log but as warning only

anta/device.py
gmuloc: in the docstring, mention that if this is not needed, it should be implemented as a `pass`. in ANTA 2.0 we will make this method abstract (so it will raise)

anta/device.py
```diff
@@ -327,6 +339,14 @@ class AsyncEOSDevice(AntaDevice):
...
         Tags for this device.
+    enable : bool
+        When True, commands are collected in privileged (enable) mode.
+    _client : asynceapi.Device
+        The underlying HTTPX-based eAPI client. Recreated by `_create_client()`.
+        Closed by `disconnect()`; automatically recreated on the next `refresh()` call.
+    _ssh_opts : SSHClientConnectionOptions
+        SSH connection configuration (host, port, credentials, host-key policy).
+        Used to establish transient SSH connections in `copy()`.
     """
```
gmuloc, line 364:
```python
    _client: asynceapi.Device
    """
    The underlying HTTPX-based eAPI client. Recreated by `_create_client()`.
    Closed by `disconnect()`; automatically recreated on the next `refresh()` call.
    """
```

anta/device.py
```diff
@@ -394,9 +414,8 @@ def __init__(  # noqa: PLR0913
             raise ValueError(message)
         self.enable = enable
         self._enable_password = enable_password
-        self._session: asynceapi.Device = asynceapi.Device(
-            host=host, port=port, username=username, password=password, proto=proto, timeout=timeout, trust_env=get_httpx_settings().trust_env
-        )
+        self._eapi_params: EapiParams = EapiParams(host=host, username=username, password=password, port=port, proto=proto, timeout=timeout)
```
gmuloc:
```suggestion
        self._eapi_params: EAPIParams = EapiParams(host=host, username=username, password=password, port=port, proto=proto, timeout=timeout)
```

anta/device.py
gmuloc: lets do `is_online=False` and same for established when you disconnect

vitthalmagadum: Added this to disconnect()

anta/device.py
vitthalmagadum: Added this to disconnect() logger.

anta/inventory/__init__.py
vitthalmagadum: Added disconnect() to AntaDevice with logger warning instead.
Please confirm.
Thanks!

anta/device.py
vitthalmagadum: Updated.

anta/device.py
vitthalmagadum: Updated. Thanks!

anta/_runner.py
```diff
@@ -302,6 +302,10 @@ async def run_with_sem(test_coro: Coroutine[Any, Any, TestResult]) -> TestResult
 
         self._log_cache_statistics(ctx)
 
+        # Disconnect from all devices after tests complete
+        with Catchtime(logger=logger, message="Disconnecting from devices"):
+            await ctx.inventory.disconnect_inventory()
```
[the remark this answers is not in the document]
gmuloc: @vitthalmagadum yes if we connect only to selected inventory that is what we should disconnect from cf https://github.com/vitthalmagadum/anta/blob/5ef23ef163f4e01a1bbdf2f3652913801f822ad9/anta/_runner.py#L347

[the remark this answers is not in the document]
gmuloc: let's mention this in the runner documentation maybe - what we connect to and what we disconnect from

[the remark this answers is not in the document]
vitthalmagadum: I was doing the same in the other PR but where Rabbit was complaining.
https://github.com/aristanetworks/anta/pull/1509#discussion_r3436882176
I'll update to selected inventory and update the doc as well.
Thanks Guillaume!

gmuloc: if we have any exception in the code before we reach here, then disconnect is not called which is kind of _not correct_ so we should rather have an encompassing try finally to disconnnect. 

Also to make this non breaking (in the sense that people having some code where they leverage the devices or check established after the call to runner, we should add a flag to control disconnecting or not (to be non breaking))

anta/_runner.py
gmuloc, line 222: we will make this True in ANTA 2.0 so you can mention it int the description
```suggestion
        dry_run: bool = False,
        disconnect: bool = False,
```

then let's add a CLI flag to decide on this..

gmuloc: pass the flag to the context for now as well so we can record

vitthalmagadum: Updated. Thanks!!

gmuloc (approved):

anta/_runner.py
```diff
@@ -79,6 +80,8 @@ class AntaRunContext:
         Manager with the final test results.
     filters: AntaRunFilters
         Provided filters to the run.
+    filtered_inventory: AntaInventory
+        Inventory matching the run device/tag filters, computed once for this run context.
```
carl-baillargeon, line 88: We are missing `dry_run` and `disconnect` attributes.

anta/cli/nrfu/__init__.py
```diff
@@ -110,6 +110,14 @@ def parse_args(self, ctx: click.Context, args: list[str]) -> list[str]:
     is_flag=True,
     default=False,
 )
+@click.option(
+    "--disconnect/--no-disconnect",
```
carl-baillargeon, line 114: Why not just `--disconnect` like `--dry-run`?

anta/device.py
```diff
@@ -311,10 +312,26 @@ async def copy(self, sources: list[Path], destination: Path, direction: Literal[
...
 
+    async def disconnect(self) -> None:
+        """Disconnect the device and close any open connections.
+
+        It is not mandatory to implement this for a valid AntaDevice subclass.
+        If disconnection logic is not needed, implement this method as ``pass``.
+        Subclasses should document the concrete resources they close.
+
+        NOTE::
+            In ANTA 2.0, this method will be made abstract and must be implemented by all subclasses.
```
carl-baillargeon:
```suggestion
        NOTE:
            In ANTA 2.0, this method will be made abstract and must be implemented by all subclasses.
```

anta/device.py
```diff
@@ -311,10 +312,26 @@ async def copy(self, sources: list[Path], destination: Path, direction: Literal[
         msg = f"copy() method has not been implemented in {self.__class__.__name__} definition"
         raise NotImplementedError(msg)
 
+    async def disconnect(self) -> None:
+        """Disconnect the device and close any open connections.
+
+        It is not mandatory to implement this for a valid AntaDevice subclass.
+        If disconnection logic is not needed, implement this method as ``pass``.
```
carl-baillargeon: We use single backticks everywhere.
```suggestion
        If disconnection logic is not needed, implement this method as `pass`.
```

anta/device.py
```diff
@@ -311,10 +312,26 @@ async def copy(self, sources: list[Path], destination: Path, direction: Literal[
...
+        Subclasses should document the concrete resources they close.
+
+        NOTE::
+            In ANTA 2.0, this method will be made abstract and must be implemented by all subclasses.
+        """
+        msg = f"disconnect() method has not been implemented in {self.__class__.__name__}"
```
carl-baillargeon: We use this structure in `copy()` above.
```suggestion
        msg = f"disconnect() method has not been implemented in {self.__class__.__name__} definition"
```

anta/device.py
```diff
@@ -327,6 +344,19 @@ class AsyncEOSDevice(AntaDevice):
         Hardware model of the device.
     tags : set[str]
         Tags for this device.
+    enable : bool
+        When True, commands are collected in privileged (enable) mode.
+    """
+
+    _client: asynceapi.Device
+    """
+    The underlying HTTPX-based eAPI client. Recreated by `_create_client()`.
```
carl-baillargeon:
```suggestion
    The underlying HTTPX-based eAPI client. Created by `_create_client()`.
```

anta/device.py
```diff
@@ -327,6 +344,19 @@ class AsyncEOSDevice(AntaDevice):
...
+    Closed by `disconnect()`; automatically recreated on the next `refresh()` call.
+    """
+    _ssh_opts: SSHClientConnectionOptions
```
carl-baillargeon, line 360: We could also add `_eapi_params` type here as well and cleanup types in `__init__`.

anta/device.py
```diff
@@ -394,9 +424,8 @@ def __init__(  # noqa: PLR0913
             raise ValueError(message)
         self.enable = enable
         self._enable_password = enable_password
-        self._session: asynceapi.Device = asynceapi.Device(
-            host=host, port=port, username=username, password=password, proto=proto, timeout=timeout, trust_env=get_httpx_settings().trust_env
-        )
+        self._eapi_params: EAPIParams = EAPIParams(host=host, username=username, password=password, port=port, proto=proto, timeout=timeout)
```
carl-baillargeon: We could call the class `EAPIClientConnectionOptions` to follow the SSH class.

gmuloc: yes let's do that

anta/cli/nrfu/__init__.py
gmuloc: well because here we are doing `--disconnect` by default :p

carl-baillargeon (approved):
LGTM if CI passes.

PR #1568, opened 2026-07-15
fix(anta.tests): Use revision 2 for show bgp neighbors commands

carl-baillargeon:
Revision 3 of `show bgp neighbors ...` commands was silently removed in EOS 4.36.X, as it should not have existed in the first place. Revision 3 will eventually be re-added as a no-op alias for revision 2, but for now we use revision 2 to avoid a non-existing revision error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BGP validation compatibility by using the correct command format when checking neighbor status, capabilities, timers, routes, and related peer details.
  * Updated exchanged-route checks to ensure advertised and received routes are evaluated reliably.

### DIFF
--- a/anta/tests/routing/bgp.py
+++ b/anta/tests/routing/bgp.py
@@ -207,7 +207,7 @@ class VerifyBGPPeersHealth(AntaTest):
     """
 
     categories: ClassVar[list[str]] = ["bgp"]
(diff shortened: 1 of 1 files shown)


gmuloc (approved):
we don't know if the revision 3 will come back as a no-op yet so better safe than sorry

PR #1584, opened 2026-07-23
fix: Invalid AntaTest top-level input types are not rejected explicitly in direct Python/API usage

### LINKED ISSUE #1580: issue: Invalid AntaTest top-level input types are not rejected explicitly in direct Python/API usage
### 📦 ANTA version

```console
anta, version 1.9.0
```

### 📝 Details

**Summary**
AntaTest._init_inputs() only handles None, dict, and AntaTest.Input. If a caller passes any other top-level type, such as str or list, the framework does not reject it explicitly. Instead, the test object can remain partially initialized and fail later with misleading runtime errors such as AttributeError.
This is primarily a Python/package API robustness bug. It is not normally hit through the standard anta nrfu catalog-file workflow, because catalog parsing validates input shape earlier.  

**Evidence**  
[models.py (line 475)](https://github.com/aristanetworks/anta/blob/main/anta/models.py#L475): _init_inputs() only handles:None

- AntaTest.Input
- dict
- Any other type falls through silently without:setting self.inputs
- marking the test result as error

**Impact**
Invalid package/API usage is not rejected at the framework boundary. Instead of a clear validation error, users get a later runtime failure that is harder to debug.

**Steps To Reproduce**

```
import asyncio
from typing import ClassVar

from anta.device import AsyncEOSDevice
from anta.models import AntaTest

class FakeTestWithInput(AntaTest):
    categories: ClassVar[list[str]] = []
    commands: ClassVar[list] = []

    class Input(AntaTest.Input):
        string: str

    @AntaTest.anta_test
    def test(self) -> None:
        self.result.is_success(self.inputs.string)

device = AsyncEOSDevice(
    host="device.example.com",
    username="admin",
    password="admin",
    name="device",
    disable_cache=True,
)

test = FakeTestWithInput(device=device, inputs="wrong-type")
asyncio.run(test.test())
print(test.result.result)
print(test.result.messages)
```
**Expected**
The framework should reject the invalid input type immediately with a clear error.

**Actual**
The test is created without valid inputs and later fails with a runtime error such as:
AttributeError: 'FakeTestWithInput' object has no attribute 'inputs'

**Proposed Fix**
Add an explicit else branch in _init_inputs() to reject unsupported top-level input types and mark the result as error with a clear message.

geetanjalimanegslab:
## Description
AntaTest._init_inputs() only handles None, dict, and AntaTest.Input. If a caller passes any other top-level type, such as str or list, the framework does not reject it explicitly. Instead, the test object can remain partially initialized and fail later with misleading runtime errors such as AttributeError.

### Fix
Add an explicit else branch in _init_inputs() to reject unsupported top-level input types and mark the result as error with a clear message. 

Fixes #1580

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for provided test inputs.
  * Unsupported input types are now handled explicitly, producing type-specific error messages.
  * Such cases consistently fail the test result and correctly surface validation issues.
* **Tests**
  * Added coverage for inputs with required fields and for additional invalid input-type scenarios (including string, list, integer, and tuple).

### DIFF
--- a/anta/models.py
+++ b/anta/models.py
@@ -477,22 +477,25 @@ class AntaTest(ABC):
 
         Any input validation error will set this test result status as 'error'.
         """
-
-        def _raise_invalid_input_type() -> None:
-            msg = f"Input should be a dict, Input instance or None, not {type(inputs).__name__}"
-            raise TypeError(msg)
-
-        try:
-            if inputs is None:
+        if inputs is None:
+            try:
                 self.inputs = self.Input()
-            elif isinstance(inputs, AntaTest.Input):
-                self.inputs = inputs
-            elif isinstance(inputs, dict):
+            except ValidationError as e:
+                message = f"{self.module}.{self.name}: Inputs are not valid\n{e}"
+                self.logger.error(message)
+                self.result.is_error(message=message)
+        elif isinstance(inputs, AntaTest.Input):
+            self.inputs = inputs
+        elif isinstance(inputs, dict):
+            try:
                 self.inputs = self.Input(**inputs)
-            else:
-                _raise_invalid_input_type()
-        except (ValidationError, TypeError) as e:
-            message = f"{self.module}.{self.name}: Inputs are not valid\n{e}"
+            except ValidationError as e:
+                message = f"{self.module}.{self.name}: Inputs are not valid\n{e}"
+                self.logger.error(message)
+                self.result.is_error(message=message)
+        else:
+            msg = f"Input should be a dict, Input instance or None, not {type(inputs).__name__}"
+            message = f"{self.module}.{self.name}: Inputs are not valid\n{msg}"
             self.logger.error(message)
             self.result.is_error(message=message)
 
--- a/anta/models.py
+++ b/anta/models.py
@@ -477,6 +477,11 @@ class AntaTest(ABC):
 
         Any input validation error will set this test result status as 'error'.
         """
+
+        def _raise_invalid_input_type() -> None:
+            msg = f"Input should be a dict, Input instance or None, not {type(inputs).__name__}"
+            raise TypeError(msg)
+
         try:
             if inputs is None:
                 self.inputs = self.Input()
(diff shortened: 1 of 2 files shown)


anta/models.py
```diff
@@ -477,14 +477,21 @@ def _init_inputs(self, inputs: dict[str, Any] | AntaTest.Input | None) -> None:
...
             if inputs is None:
                 self.inputs = self.Input()
             elif isinstance(inputs, AntaTest.Input):
                 self.inputs = inputs
             elif isinstance(inputs, dict):
                 self.inputs = self.Input(**inputs)
-        except ValidationError as e:
+            else:
+                _raise_invalid_input_type()
+        except (ValidationError, TypeError) as e:
```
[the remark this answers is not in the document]
geetanjalimanegslab: addressed

tests/units/test_models.py
```diff
@@ -297,8 +297,29 @@ class FakeTestWithMissingTest(AntaTest):
...
+        """Test function."""
+        self.result.is_success()
+
+
 ANTATEST_DATA: dict[tuple[type[AntaTest], str], Any] = {
     (FakeTest, "no input"): {"inputs": None, "expected": {"__init__": {"result": "unset"}, "test": {"result": "success"}}},
+    (FakeTest, "Input instance with no fields"): {
+        "inputs": FakeTest.Input(),
+        "expected": {"__init__": {"result": "unset"}, "test": {"result": "success"}},
+    },
```
vitthalmagadum, line 322: This is not a right way, right?
Here we should have something to reject such things.
For example, One my test has 
```
class Input(AntaTest.Input):
        abc: list[ABC]
```
Allows us, initialize the instance with inputs={"abc": []}, and the loop processes nothing and the test reports success. 
@gmuloc Do we need to address this on input level or can we have something here?

gmuloc: this should be done at input level to make sure we don't accept empty list I think what is your concrete example?
I am not sure I understand all your concern properly here can you elaborate?

vitthalmagadum: Right, So we should have something like in input models only:
```
class Input(AntaTest.Input):
        abc: list[ABC] = Field(min_length=1)
```
But in that case we need to revisit all the models.
Resolving the comments thanks @gmuloc

vitthalmagadum (approved):

gmuloc (approved):

PR #1585, opened 2026-07-23
fix: AntaDevice mutates shared tag sets, causing cross-device tag contamination for inventory networks and ranges

### LINKED ISSUE #1577: issue: AntaDevice mutates shared tag sets, causing cross-device tag contamination for inventory networks and ranges
### 📦 ANTA version

```console
anta version 1.9
```

### 📝 Details

**Summary:**
AntaDevice mutates the tags set object passed into the constructor by adding self.name directly to it. This becomes a real ANTA bug because inventory parsing reuses the same tags set when expanding a single networks or ranges entry into multiple devices.
As a result, devices created from the same inventory block can inherit each other’s names as tags.

**Evidence:**  
[device.py (line 168)](https://github.com/aristanetworks/anta/blob/main/anta/device.py#L168) stores the incoming tags set directly and mutates it with self.tags.add(self.name)
[inventory/__init__.py (line 193)](https://github.com/aristanetworks/anta/blob/main/anta/inventory/__init__.py#L193) passes network.tags directly into each AsyncEOSDevice
[inventory/__init__.py (line 244)](https://github.com/aristanetworks/anta/blob/main/anta/inventory/__init__.py#L244) passes range_def.tags directly into each AsyncEOSDevice

**Impact:**
Devices created from the same network or range can inherit each other’s names as tags, which can lead to incorrect device filtering, matching, and test selection.

**Steps to Reproduce:**
Use either ranges or networks.

**Example using ranges:**
```
anta_inventory:
  ranges:
    - start: 10.0.0.1
      end: 10.0.0.2
      tags: ["spine"]
```

```
from pathlib import Path
from anta.inventory import AntaInventory

inventory = AntaInventory.parse(
    filename=Path("repro_inventory.yml"),
    username="admin",
    password="admin",
)

for device in inventory.devices:
    print(device.name, sorted(device.tags))
```

**Expected**  
```
10.0.0.1 ['10.0.0.1', 'spine']
10.0.0.2 ['10.0.0.2', 'spine']
```
**Actual**  
```
10.0.0.1 ['10.0.0.1', '10.0.0.2', 'spine']
10.0.0.2 ['10.0.0.1', '10.0.0.2', 'spine']
```
The same issue also occurs with networks, for example:
```
anta_inventory:
  networks:
    - network: 10.0.0.0/31
      tags: ["spine"]
```

**Proposed fix**
Make a defensive copy before mutating:
```
self.tags = set(tags) if tags is not None else set()
self.tags.add(self.name)
```

geetanjalimanegslab:
## Description
 AntaDevice mutates shared tag sets, causing cross-device tag contamination for inventory networks and ranges

Fixes #1577 

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Device tags are now isolated when creating devices from shared tag collections.
  * Device names are consistently included in each device’s tags.
  * Devices created without tags now receive only their device name as a tag.

### DIFF
--- a/anta/device.py
+++ b/anta/device.py
@@ -165,7 +165,7 @@ class AntaDevice(ABC):
         """
         self.name: str = name
         self.hw_model: str | None = None
-        self.tags: set[str] = tags if tags is not None else set()
+        self.tags: set[str] = set(tags) if tags is not None else set()
         # A device always has its own name as tag
         self.tags.add(self.name)
         self.is_online: bool = False
--- a/anta/device.py
+++ b/anta/device.py
@@ -165,7 +165,7 @@ class AntaDevice(ABC):
         """
         self.name: str = name
         self.hw_model: str | None = None
-        self.tags: set[str] = set(tags) if tags is not None else set()
+        self.tags: set[str] = tags.copy() if tags is not None else set()
         # A device always has its own name as tag
         self.tags.add(self.name)
         self.is_online: bool = False
--- a/tests/units/test_device.py
+++ b/tests/units/test_device.py
@@ -861,3 +861,73 @@ class TestAsyncEOSDevice:
         cmd = AntaCommand(command="show version")
         with pytest.raises(RuntimeError, match="httpx client is closed"):
             await async_device._collect(cmd)
+
+    def test_tags_set_not_mutated(self) -> None:
+        """Verify that passing a tags set does not mutate the original set."""
+        shared_tags = {"tag1", "tag2"}
+        original_tags = shared_tags.copy()
+
+        AsyncEOSDevice(
+            host="42.42.42.42",
+            username="anta",
+            password="anta",
+            name="device1",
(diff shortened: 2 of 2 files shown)


vitthalmagadum (approved):

anta/device.py
```diff
@@ -165,7 +165,7 @@ def __init__(self, name: str, tags: set[str] | None = None, *, disable_cache: bo
         """
         self.name: str = name
         self.hw_model: str | None = None
-        self.tags: set[str] = tags if tags is not None else set()
+        self.tags: set[str] = set(tags) if tags is not None else set()
```
gmuloc: given that we should only receive `set` or None, tags.copy() should be a little faster as it does not invoke the `set` constructor (already some guarantees)
```suggestion
        self.tags: set[str] = tags.copy() if tags is not None else set()
```

gmuloc (approved):

PR #1587, opened 2026-07-24
fix: read console authentication methods from EOS output

### LINKED ISSUE #1586: bug: VerifyAuthenMethods can never succeed for `type: login` against current EOS
**Intended for:** https://github.com/aristanetworks/anta/issues
**Affected package:** `anta`
**Affected version:** `1.9.0` (currently the latest release on PyPI as of Jul 2026 — checked, no newer release available)
**Severity:** Functional bug — `anta.tests.aaa.VerifyAuthenMethods` with `types: [login, ...]` fails unconditionally, regardless of the device's actual AAA configurati
on.

## Summary

`VerifyAuthenMethods` (`anta/tests/aaa.py`, class starting at line 159) contains
a special case for `auth_type == "login"` that requires a JSON key literally
named `"login"` inside the `loginAuthenMethods` dict returned by
`show aaa methods authentication`. On real EOS (verified against a
`cEOS64-lab-4.36.1F` container, and matching Arista's own CLI output format),
that dict only ever contains `default` and, if a console-specific method list
is configured, `console` — **never** `login`. As a result, any catalog entry
with `types: [login, ...]` fails immediately with
`"AAA authentication methods are not configured for login console"`, even
when the device's actual login authentication is correctly configured and
matches the expected `methods`.

# Reproduction

Device config (rendered via Arista AVD, `aaa_settings.authentication.login`):

```
aaa authentication login console local
```

(`login default` is not rendered as an explicit line because it already
matches the EOS built-in default — this is normal AVD/EOS behavior, not a
misconfiguration.)

Real JSON output of `show aaa methods authentication | json` from that device:

```json
{
    "loginAuthenMethods": {
        "default": {
            "methods": ["local"]
        },
        "console": {
            "methods": ["local"]
        }
    },
    "enableAuthenMethods": {
        "default": {
            "methods": ["local"]
        }
    },
    "dot1xAuthenMethods": {
        "default": {
            "methods": []
        }
    }
}
```

atalog entry:

```yaml
anta.tests.aaa:
  - VerifyAuthenMethods:
      methods:
        - local
      types:
        - login
```

Result:

```
FAILURE
AAA authentication methods are not configured for login console
```

This fails **regardless of `methods`, regardless of whether `console` is
configured, and regardless of whether `default` actually matches** — the test
never gets far enough to compare anything, because the guard clause itself is
unsatisfiable.

## Root cause

`anta/tests/aaa.py`, `VerifyAuthenMethods.test`:

```python
for k, v in command_output.items():
    auth_type = k.replace("AuthenMethods", "")
    if auth_type not in self.inputs.types:
        continue
    if auth_type == "login":
        if "login" not in v:
            self.result.is_failure("AAA authentication methods are not configured for login console")
            return
        if v["login"]["methods"] != self.inputs.methods:
            self.result.is_failure(f"AAA authentication methods {', '.join(self.inputs.methods)} are not matching for login console")
            return
    not_matching.extend(auth_type for methods in v.values() if methods["methods"] != self.inputs.methods)
```

(lines 204–208 in the installed 1.9.0 distribution)
`v` here is `command_output["loginAuthenMethods"]`, whose real sub-keys are
`default`/`console` — never `login`. Given the failure message explicitly
says `"...for login console"`, the intent looks like it was meant to index
`v["console"]`, not `v["login"]` — a copy-paste/naming mix-up between the
JSON key (`login`) used as the dict-of-dicts name (`loginAuthenMethods`) and
the sub-key that should have been checked (`console`).

## Proposed fix

Replace the hardcoded `"login"` sub-key lookup with `"console"`:

```python
if auth_type == "login":
    if "console" not in v:
        self.result.is_failure("AAA authentication methods are not configured for login console")
        return
    if v["console"]["methods"] != self.inputs.methods:
        self.result.is_failure(f"AAA authentication methods {', '.join(self.inputs.methods)} are not matching for login console")
        return
```

Alternatively (and possibly preferable, since not every deployment configures
a console-specific override — ours didn't, until we added one specifically
to work around this bug): drop the mandatory-`console`-key requirement
entirely and let the existing generic `not_matching` loop at the bottom
(which already iterates `v.values()` and compares each sub-dict's `methods`)
handle `login` the same way it already handles `enable`/`dot1x`. That would
let `types: [login]` correctly validate whichever of `default`/`console` are
actually present, instead of requiring an explicit console override just to
make the test runnable at all.

## Environment

- `anta` version: 1.9.0
- EOS: `cEOS64-lab-4.36.1F-48405728.4361F (engineering build)`
- Reproduced via `arista.avd`-generated configuration (Ansible collection),
  but the bug is purely in ANTA's JSON-key handling and is independent of how
  the config was produced.

Sanjays2402:
## Description

`VerifyAuthenMethods` looked for a nonexistent `login` entry inside EOS `loginAuthenMethods` output, causing valid login checks to fail before comparing methods. Read the actual `console` entry and align the AAA fixtures with current EOS output.

Fixes #1586

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected validation of console authentication methods for login authentication checks.
  * Updated authentication test scenarios to accurately reflect console-based method configuration.

### DIFF
--- a/anta/tests/aaa.py
+++ b/anta/tests/aaa.py
@@ -202,10 +202,10 @@ class VerifyAuthenMethods(AntaTest):
                 # We do not need to verify this accounting type
                 continue
             if auth_type == "login":
-                if "login" not in v:
+                if "console" not in v:
                     self.result.is_failure("AAA authentication methods are not configured for login console")
                     return
-                if v["login"]["methods"] != self.inputs.methods:
+                if v["console"]["methods"] != self.inputs.methods:
                     self.result.is_failure(f"AAA authentication methods {', '.join(self.inputs.methods)} are not matching for login console")
                     return
             not_matching.extend(auth_type for methods in v.values() if methods["methods"] != self.inputs.methods)
--- a/tests/units/anta_tests/test_aaa.py
+++ b/tests/units/anta_tests/test_aaa.py
@@ -158,7 +158,7 @@ DATA: AntaUnitTestData = {
     (VerifyAuthenMethods, "success-login-enable"): {
         "eos_data": [
             {
-                "loginAuthenMethods": {"default": {"methods": ["group tacacs+", "local"]}, "login": {"methods": ["group tacacs+", "local"]}},
+                "loginAuthenMethods": {"default": {"methods": ["group tacacs+", "local"]}, "console": {"methods": ["group tacacs+", "local"]}},
                 "enableAuthenMethods": {"default": {"methods": ["group tacacs+", "local"]}},
                 "dot1xAuthenMethods": {"default": {"methods": ["group radius"]}},
             }
@@ -169,7 +169,7 @@ DATA: AntaUnitTestData = {
     (VerifyAuthenMethods, "success-dot1x"): {
         "eos_data": [
             {
-                "loginAuthenMethods": {"default": {"methods": ["group tacacs+", "local"]}, "login": {"methods": ["group tacacs+", "local"]}},
+                "loginAuthenMethods": {"default": {"methods": ["group tacacs+", "local"]}, "console": {"methods": ["group tacacs+", "local"]}},
                 "enableAuthenMethods": {"default": {"methods": ["group tacacs+", "local"]}},
                 "dot1xAuthenMethods": {"default": {"methods": ["group radius"]}},
             }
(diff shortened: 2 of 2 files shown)


gmuloc (changes requested):
please see my comment on the issue. putting this on hold until I get more info from dev

gmuloc:
using #1594 to fix it. Thanks a lot for reporting

PR #1633, opened 2026-08-26
feat(anta.tests): Add test for Arista Security Advisory 0117

carl-baillargeon:
## Description

See title.

Fixes # (issue id)

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added validation for Arista Security Advisory SA-117 (CVE-2025-0936).
  - Checks EOS versions, gNMI accounting, and OpenConfig trace settings to identify potentially affected devices.
  - Added advisory configuration to the example test suite.

- **Tests**
  - Added coverage for EOS version parsing, advisory evaluation, configuration scenarios, and invalid or incomplete inputs.
  - Added success and failure cases for SA-117 detection.


anta/tests/advisories/sa_117.py
```diff
@@ -0,0 +1,131 @@
...
+
+        if not require_affected_version(self.result, messages, version_output, AFFECTED_VERSION_MATRIX):
+            return
+
+        accounting_enabled = _evaluate_gnmi_accounting_enabled(gnmi_output)
+        risky_trace_configured = _evaluate_risky_trace_configuration(running_config_output)
+        if accounting_enabled is True:
+            messages.append("OpenConfig gNMI has accounting requests enabled.")
+        elif risky_trace_configured is True:
+            messages.append("OpenConfig tracing includes one of the risky selectors from the advisory.")
```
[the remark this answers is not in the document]
carl-baillargeon: Good catch. Also, if a transport is configured with both accounting and tracing enabled, we should get both failure messages in the result.

anta/tests/advisories/sa_117.py
```diff
@@ -0,0 +1,131 @@
...
+        "On affected platforms running Arista EOS with a gNMI transport enabled, running the gNOI "
+        "File TransferToRemote RPC with credentials for a remote server may cause these remote-server "
+        "credentials to be logged or accounted on the local EOS device or possibly on other remote "
+        "accounting servers (i.e. TACACS, RADIUS, etc)."
+    ),
+}
+
+
+class VerifySA117(AntaTest):
+    """Verify that the device is not exposed to Arista Security Advisory 0117 (CVE-2025-0936).
```
carl-baillargeon, line 106: When looking at the SA, it looks like mitigation options 2 and 3 are not evaluated in the current code:

- `OCGNOIFileToggle=0`
- gNSI Authz blocking `/gnoi.file.File/TransferToRemote`

Should we add these checks to avoid failures on mitigated devices?

@gmuloc @ClausHolbechArista @ankudinov

carl-baillargeon: A user could decide to use these workarounds but the test would still fail after applying them.

anta/tests/advisories/_version.py
```diff
@@ -0,0 +1,125 @@
+# Copyright (c) 2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+"""Affected EOS version helpers for security advisory tests."""
```
gmuloc, line 4: I would move this outside of advisories in tools (which may need to become a folder again and keeping backward compatibility)

anta/tests/advisories/sa_117.py
ClausHolbechArista: These policies can be changed adhoc, so I chose the careful approach. Petr and I discuss a possible input knob on the tests to declare "mitigated"

anta/tests/advisories/sa_117.py
```diff
@@ -0,0 +1,131 @@
...
+
+from anta.models import AntaCommand, AntaTemplate, AntaTest
+from anta.tests.advisories._version import VersionRule, require_affected_version
+
+RISKY_TRACE_SELECTORS = (
+    "service/9",
+    "interceptor/9",
+    "transport_socketcli/9",
+)
+AFFECTED_VERSION_MATRIX: tuple[VersionRule, ...] = (
```
gmuloc, line 19: this can move to the subclass of AntaTest in a next PR

anta/tests/advisories/sa_117.py
```diff
@@ -0,0 +1,131 @@
...
+        "credentials to be logged or accounted on the local EOS device or possibly on other remote "
+        "accounting servers (i.e. TACACS, RADIUS, etc)."
+    ),
+}
+
+
+class VerifySA117(AntaTest):
+    """Verify that the device is not exposed to Arista Security Advisory 0117 (CVE-2025-0936).
+
+    Expected Results
```
gmuloc, line 112: maybe we put a note in there about mitigation not being taken into account for now?

tests/benchmark/test_anta.py
```diff
@@ -88,5 +89,6 @@ def _() -> None:
         "---------------------------------------"
     )
     logger.info(bench_info)
-    assert results.get_total_results({AntaTestStatus.ERROR}) == 0
+    expected_errors = len(inventory) * anta_mock_env.expected_status_counts[AntaTestStatus.ERROR]
```
gmuloc: hmmm we had a plan to make sure we would never report error from tests that's why it was there

gmuloc: ERROR is supposed to be only "there is a problem in the logic"

anta/tests/advisories/sa_117.py
```diff
@@ -0,0 +1,131 @@
...
+
+        accounting_enabled = _evaluate_gnmi_accounting_enabled(gnmi_output)
+        risky_trace_configured = _evaluate_risky_trace_configuration(running_config_output)
+        if accounting_enabled is True:
+            messages.append("OpenConfig gNMI has accounting requests enabled.")
+        elif risky_trace_configured is True:
+            messages.append("OpenConfig tracing includes one of the risky selectors from the advisory.")
+        elif accounting_enabled is None or risky_trace_configured is None:
+            messages.append("The gNMI accounting or OpenConfig trace configuration could not be determined from the available EOS command output.")
+            self.result.is_error("\n".join(messages))
```
gmuloc, line 162: let's use inconclusive here? or I think maybe Claus was saying it is actually an error but it should be a failure really to keep ANTA philosophy

tests/units/anta_tests/advisories/test_version.py
```diff
@@ -0,0 +1,97 @@
+# Copyright (c) 2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+"""Unit tests for security advisory EOS version helpers."""
```
gmuloc, line 4: then lets move the tests to tools as well

tests/units/anta_tests/__init__.py
```diff
@@ -36,11 +36,10 @@ class AtomicResult(TypedDict):
 class UnitTestResult(TypedDict):
     """Expected result of a unit test of an AntaTest subclass.
 
-    For our AntaTest unit tests we expect only success, failure or skipped.
-    Never unset nor error.
```
gmuloc, line 40: I don't think we want that really

gmuloc: also this would need to be reflected in doc if we do this

anta/tests/advisories/sa_117.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @AntaTest.anta_test
+    def test(self) -> None:
+        """Fail when the exposure signals and every applicable affected condition are present."""
+        gnmi_output = self.instance_commands[0].json_output
+        running_config_output = self.instance_commands[1].json_output
+        version_output = self.instance_commands[2].json_output
+        messages: list[str] = []
+
+        if not require_affected_version(self.result, messages, version_output, AFFECTED_VERSION_MATRIX):
```
gmuloc, line 141: as per the advisory, should we check the affected platforms as well? (maybe as we are doing in Field Notice?)

anta/tests/advisories/__init__.py
gmuloc, line 1: need to add doc

anta/tests/advisories/sa_117.py
gmuloc: it affects all platforms

tests/units/anta_tests/advisories/test_sa_117.py
```diff
@@ -0,0 +1,130 @@
...
+        },
+    },
+    (VerifySA117, "success-excluded-version-suffix"): {
+        "eos_data": [{}, {}, {"version": "4.33.1FX-wbb"}],
+        "expected": {
+            "result": AntaTestStatus.SUCCESS,
+            "messages": ["The EOS version '4.33.1FX-wbb' is not affected by this advisory."],
+        },
+    },
+    (VerifySA117, "error-unknown-configuration"): {
```
gmuloc: lets comment these error tests for now and revisit at alater point

anta/tests/advisories/sa_117.py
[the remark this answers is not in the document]
carl-baillargeon: Fixed your initial comment in b18c775ef5547644639c543eaf6b5e4ca85b7d95. Also, please remove the learning for both failure messages for now. We will revisit later.

tests/units/anta_tests/advisories/test_sa_117.py
carl-baillargeon: Removed them.

tests/units/anta_tests/__init__.py
carl-baillargeon: Removed the error unit tests for now.

anta/tests/advisories/sa_117.py
carl-baillargeon: Keeping as error for now we can revisit later.

tests/benchmark/test_anta.py
carl-baillargeon: Removed the error unit tests for now.

anta/tests/advisories/__init__.py
carl-baillargeon: Done.

gmuloc (approved):
as discussed, this is all preview so we will get them inside

PR #1733, opened 2026-09-10
fix(anta): apply connectivity timeout to session login

carl-baillargeon:
## Description

When `--use-session-auth` is enabled, the initial `POST /login` request in `EapiSessionAuth` did not set a connect timeout. On unreachable inventory hosts, inventory refresh could block for ~2 minutes (OS TCP retry window) instead of failing fast like basic-auth refresh via `check_api_endpoint()`.

This change applies the same 5-second connectivity timeout to session login and centralizes the value as `EAPI_CONNECTIVITY_TIMEOUT`.

Fixes # (issue id)

## Checklist

### DIFF
--- a/asynceapi/_auth.py
+++ b/asynceapi/_auth.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING
 
 import httpx
 
+from ._constants import EAPI_CONNECTIVITY_TIMEOUT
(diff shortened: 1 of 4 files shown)


gmuloc (approved):
tested and works better a follow up will be to mention this in our doc as right now we have an ambiguity regarding --timeout not impacting the login timeout (with and without session)

PR #1740, opened 2026-09-11
refactor(anta): Advisory - Migrate to nominal typed facts

ClausHolbechArista:
Migrate the advisory framework from generic collected values to nominal, strongly typed fact instances, and adopt the nested typed `Facts` container across every advisory.

## Summary

- make each available observation contain an instance of its concrete fact-definition class
- annotate advisory fields as `Fact[ConcreteFact]`, preserving the concrete type through collection and assessment
- migrate every production advisory to `@facts_dataclass` and `self.Facts.collect(self)`
- move definition caching, command derivation, ordered command slicing, and bulk collection into `FactsBase`
- derive `AvailableFact.definition` from `type(value)` so the value and definition cannot disagree
- create available observations from nominal instances with `value.available(source)`
- retain explicit unavailable observations with their definition, problem, source, and validated contradictory observations
- remove the transitional `required_facts`, `_AntaAdvisoryTest.fact()`, and one-at-a-time advisory collection APIs
- remove legacy generic value models and their dual-runtime handling

## Authoring pattern

```python
@facts_dataclass
class Facts(FactsBase):
    version: Fact[EosVersionFact] = fact_field(EosVersionFact)
    bfd_authentication: Fact[BfdAuthenticationFact] = fact_field(BfdAuthenticationFact)
```

```python
facts = self.Facts.collect(self)
finding = _assess_sa154(facts.version, facts.bfd_authentication)
```

Concrete fact definitions are frozen dataclasses containing their normalized payload. Identity and rendering data such as `key`, `label`, `commands`, and feature identity remain class-level metadata on each concrete definition.

## Static and runtime boundary

`fact_field()` is statically typed as returning `Fact[T]`, while returning a required `dataclasses.Field` at runtime. `@facts_dataclass` registers it as a PEP 681 field specifier so Pyright connects `FactDefinition[T]` to the annotated `Fact[T]` field and understands the generated constructor.

The decorator applies the frozen, slotted dataclass transformation, validates declarations once at class definition, and stores an immutable definition mapping. Runtime validation deliberately does not inspect type annotations. It checks that:

- every field has exactly the expected `definition` metadata
- every metadata value is a `FactDefinition` subclass
- every field participates in the generated initializer
- definitions are unique within the container
- the container declares at least one fact

An undecorated `FactsBase` subclass is rejected when its definitions are consumed. The erased generic relationship remains the type checker's responsibility; runtime checks retain the command-count and command-identity guarantees needed to safely associate collected responses with definitions.

## Nominal observation model

- `AvailableFact[T]` contains a concrete nominal `T` and derives its definition from that value's runtime class.
- `UnavailableFact[T]` retains the requested definition and validates any contradictory observations against it.
- `EosVersionFact` retains equality and hash compatibility with normalized `EOSVersion` values while remaining nominally typed.
- Every concrete production fact directly annotates `key` and `label`; command and feature metadata are likewise annotated on the concrete class, avoiding metadata hidden several MRO levels away from static analysis.

## Scope

This now covers all production advisory fact definitions and advisories from SA117 through SA178 currently present in the repository. Assessment, finding projection, shared test builders, and unit tests have been updated to consume nominal fact values consistently.

## Validation

- focused migrated tests: `79 passed`, `42 subtests passed`
- complete advisory/facts unit suites: `1010 passed`, `377 subtests passed`
- Ruff format and lint: clean
- Pyright: no project diagnostics in the changed focused files; only unresolved local `pytest` imports caused by the local interpreter environment
- annotation audit: every concrete fact has direct identity annotations and immediate payload annotations
- migration completion searches and `git diff --check`: clean

GitHub CI is rerunning against the complete migration.

## Summary by CodeRabbit

- **New Features**
  - Added typed, declarative fact containers for advisory data collection.
  - Added validated fact models for configuration, features, mitigations, software versions, platforms, and identities.
  - Advisory commands are now derived automatically from declared facts.

- **Breaking Changes**
  - Removed the legacy `required_facts`, `PendingFact`, and manual fact-collection APIs.
  - Removed legacy indicator value support; integrations must use the new fact models.

### DIFF
--- a/anta/_advisory/base.py
+++ b/anta/_advisory/base.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import sys
 from typing import TYPE_CHECKING, Any, ClassVar, TypeVar
 
+from anta._advisory.facts.models import PendingFact
 from anta._advisory.models import _AdvisoryMetadata
 from anta._advisory.results import _AdvisoryTestResult
 from anta.models import AntaCommand, AntaTemplate, AntaTest, _description_from_docstring
@@ -18,7 +19,7 @@ else:
     from typing_extensions import override
 
 if TYPE_CHECKING:
-    from anta._advisory.facts.models import Fact, FactDefinition
+    from anta._advisory.facts.models import CollectedFact, Fact, FactDefinition
 
 T = TypeVar("T")
 
@@ -51,7 +52,7 @@ class _AntaAdvisoryTest(AntaTest):
         """Derive commands, set subclass identity, and validate advisory attributes."""
         has_own_name = "name" in cls.__dict__
         has_own_description = "description" in cls.__dict__
-        required_facts = cls.__dict__.get("required_facts", ())
+        required_facts = cls._required_facts_for_subclass()
         if required_facts:
             if "commands" in cls.__dict__:
                 msg = f"Class {cls.__module__}.{cls.__name__} cannot define both 'required_facts' and 'commands'"
@@ -78,11 +79,27 @@ class _AntaAdvisoryTest(AntaTest):
             raise AttributeError(msg)
 
     @classmethod
+    def _required_facts_for_subclass(cls) -> tuple[type[FactDefinition[Any]], ...]:
+        """Return facts declared through either the legacy tuple or typed fields."""
+        required_facts = cls.__dict__.get("required_facts", ())
+        if not (facts_type := cls.__dict__.get("Facts")):
+            return required_facts
+        if required_facts:
+            msg = f"Class {cls.__module__}.{cls.__name__} cannot define both 'Facts' and 'required_facts'"
+            raise AttributeError(msg)
+        declared_facts = vars(facts_type())
+        if not declared_facts or any(not isinstance(fact, PendingFact) for fact in declared_facts.values()):
+            msg = f"Class {cls.__module__}.{cls.__name__}.Facts must declare one or more PendingFact defaults"
+            raise TypeError(msg)
+        cls.required_facts = tuple(fact.definition for fact in declared_facts.values())
+        return cls.required_facts
+
+    @classmethod
     def _commands_from_required_facts(cls, required_facts: tuple[type[FactDefinition[Any]], ...]) -> list[AntaCommand | AntaTemplate]:
         """Return the commands needed by the required facts in declaration order."""
         return [command for definition in required_facts for command in definition.required_commands()]
 
-    def fact(self, definition: type[FactDefinition[T]]) -> Fact[T]:
+    def fact(self, definition: type[FactDefinition[T]]) -> CollectedFact[T]:
         """Derive one required fact from device metadata or collected command data."""
         command_offset = 0
         for candidate in self.required_facts:
@@ -94,3 +111,9 @@ class _AntaAdvisoryTest(AntaTest):
 
         msg = f"Fact '{definition.key}' is not listed in required_facts for {self.__class__.__name__}"
         raise ValueError(msg)
+
+    def collect_fact(self, fact: Fact[T]) -> CollectedFact[T]:
+        """Collect a pending fact, or return an already collected fact unchanged."""
+        if isinstance(fact, PendingFact):
+            return self.fact(fact.definition)
+        return fact
--- a/anta/_advisory/facts/eos.py
+++ b/anta/_advisory/facts/eos.py
@@ -8,8 +8,8 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from anta._advisory.facts.models import (
+    CollectedFact,
     CommandsFactDefinition,
-    Fact,
     FactDefinition,
     FactProblemKind,
     FactSource,
@@ -32,7 +32,7 @@ class EosVersionFact(FactDefinition[EOSVersion]):
     label = "EOS version"
 
     @classmethod
-    def derive(cls, device: AntaDevice, commands: tuple[AntaCommand, ...] = ()) -> Fact[EOSVersion]:
+    def derive(cls, device: AntaDevice, commands: tuple[AntaCommand, ...] = ()) -> CollectedFact[EOSVersion]:
         """Normalize the device version into an EOS version fact."""
         _ = commands
         source = FactSource("device metadata", FactSourceKind.DEVICE_METADATA)
@@ -53,7 +53,7 @@ class SecureBootFact(CommandsFactDefinition[FeatureValue]):
     commands = (AntaCommand(command="show boot", revision=1),)
 
     @classmethod
-    def parse(cls, commands: tuple[AntaCommand, ...]) -> Fact[FeatureValue]:
+    def parse(cls, commands: tuple[AntaCommand, ...]) -> CollectedFact[FeatureValue]:
         """Normalize the collected command output into a Secure Boot fact.
 
         The structured fields prove the platform-support and configuration prerequisites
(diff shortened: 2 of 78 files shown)


anta/_advisory/facts/models.py
```diff
@@ -114,6 +124,105 @@ def __post_init__(self) -> None:
...
+
+        Looking directly in ``cls.__dict__`` prevents a subclass from accidentally
+        reusing a cache created for another fact-container class.
+        """
+        if "_definitions" not in cls.__dict__:
+            cls._validate_definitions()
+        return cls._definitions
+
+    @classmethod
+    def _validate_definitions(cls) -> None:
```
gmuloc: Should we reject duplicate definitions? _AntaAdvisoryTest.fact() resolves by definition identity and always returns the first occurrence, so duplicate fact_field(XFact) declarations would silently reuse the first command slice and leave the second unused.

anta/_advisory/facts/models.py
```diff
@@ -191,9 +300,17 @@ class FeatureState(str, Enum):
     UNSUPPORTED = "not supported"
 
 
+@dataclass(frozen=True, slots=True)
+class FeatureFact:
+    """Common state and class-level identity for a nominal feature fact."""
+
+    feature: ClassVar[FeatureRef]
```
gmuloc, line 374: should we make sure feature is always defined? we are using it later in Projection

anta/_advisory/facts/models.py
```diff
@@ -85,22 +95,22 @@ def required_commands(cls) -> tuple[AntaCommand, ...]:
 
 
 @dataclass(frozen=True, slots=True)
-class AvailableFact(Generic[T]):
+class AvailableFact(Generic[T_co]):
     """A typed fact whose normalized value is known."""
 
-    definition: type[FactDefinition[T]]
-    value: T
+    definition: type[FactDefinition[T_co]]
```
gmuloc: right now this opens the possibility to create weird objects in code so we will need a follow up PR to clean this up


The problem is that definition exposes a method that accepts a value:

```
  FactDefinition.available(value: T, source)
```

Consider:

```
  def rebuild(
      fact: AvailableFact[FeatureFact],
  ) -> AvailableFact[FeatureFact]:
      return fact.definition.available(
          OtherFeatureFact(FeatureState.ENABLED),
          source,
      )

  bfd: AvailableFact[BfdAuthenticationFact] = ...
  bad = rebuild(bfd)
```

  Pyright accepts the call because, inside rebuild, it sees:

```
  fact.definition: type[FactDefinition[FeatureFact]]
```
  At runtime, however, the actual definition remains `BfdAuthenticationFact`. The result becomes:

```
  AvailableFact(
      definition=BfdAuthenticationFact,
      value=OtherFeatureFact(...),
  )
```

ClausHolbechArista: This is already mentioned in the pr description as well as the todo on the available() method that it should move to the instance instead of the class. Please focus on the areas specifically mentioned as reviewable for now.

anta/_advisory/facts/models.py
ClausHolbechArista: Well even if you grab the "wrong" command it won't make a difference since it is the same command invocation on teh device. I was thinking that this should wait until we build a different hierarchy where the facts are somehow involved more in the command infrastructure.
That said can absolutely validate for unique facts.

anta/_advisory/facts/models.py
ClausHolbechArista: I am not sure I follow. How would feature not be defined if it is given here?

anta/_advisory/facts/models.py
gmuloc: yes, this is merely an example of what happens today so we need to fix this before scaling. Can we update the PR description and remove PendingFact as this is not accurate anymore?

anta/_advisory/facts/models.py
ClausHolbechArista: merged your pr so now we have duplicate check

anta/_advisory/facts/models.py
ClausHolbechArista: got it now. Fixing

gmuloc (commented):
Ok to continue on this track

anta/_advisory/base.py
```diff
@@ -77,6 +78,22 @@ def __init_subclass__(cls) -> None:
...
+    @classmethod
+    def _required_facts_for_subclass(cls) -> tuple[type[FactDefinition[Any]], ...]:
+        """Return facts declared through either the legacy tuple or typed fields."""
+        required_facts = cls.__dict__.get("required_facts", ())
+        if not (facts_type := cls.__dict__.get("Facts")):
+            return required_facts
+        if required_facts:
+            msg = f"Class {cls.__module__}.{cls.__name__} cannot define both 'Facts' and 'required_facts'"
+            raise AttributeError(msg)
+        declared_facts = vars(facts_type())
```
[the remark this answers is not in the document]
ClausHolbechArista: this has been fixed for a while. please confirm.

anta/_advisory/facts/eos.py
```diff
@@ -25,14 +27,43 @@
...
-    key = "eos.version"
-    label = "EOS version"
+    key: ClassVar[str] = "eos.version"
+    label: ClassVar[str] = "EOS version"
 
     @classmethod
-    def derive(cls, device: AntaDevice, commands: tuple[AntaCommand, ...] = ()) -> Fact[EOSVersion]:
+    def from_version(cls, version: EOSVersion) -> EosVersionFact:
+        """Create the nominal fact value from an already normalized EOS version."""
+        return cls(version.major, version.minor, version.patch, version.suffix, version.hotfix)
```
gmuloc: hardening to avoid being dependent of the EOSVersion dataclass order

```suggestion
        return cls(major=version.major, minor=version.minor, patch=version.patch, suffix=version.suffix, hotfix=version.hotfix)
```

gmuloc: should do the same for platforms and probably add some extra unit tests to enforce the contracts.


I asked my agent to suggest tests to enforce the contract

  In tests/units/_advisory/facts/test_eos.py:

  1. test_eos_version_fact_from_version_preserves_all_fields
      - Use distinct suffix and non-zero hotfix.
      - Assert every field, str(), and to_dict() match the source.

  2. test_eos_version_fact_preserves_cross_type_ordering
      - Compare EosVersionFact < EOSVersion.
      - Compare EOSVersion < EosVersionFact.

  3. test_eos_version_fact_is_interchangeable_as_mapping_key
      - Store with one type and retrieve with the other, in both directions.

Keep the existing bidirectional equality and hash test.

  In tests/units/_advisory/facts/test_platform.py:

  4. test_platform_identity_fact_from_identity_preserves_all_fields
      - Cover model, type, modules, and platform_families.
      - Assert fact.to_dict() == platform.to_dict().

  5. test_platform_identity_fact_derive_returns_nominal_value
      - Assert the result is AvailableFact.
      - Assert definition is PlatformIdentityFact.
      - Assert value is PlatformIdentityFact.
      - Assert all source fields are preserved.

  6. test_switch_card_identity_fact_from_identity_preserves_all_fields
      - Cover model, role, slot, and platform_families.
      - Assert serialization parity with the source component.

  7. Extend test_switch_card_identity
      - Assert definition is SwitchCardIdentityFact.
      - Assert the value is nominal.
      - Assert slot and platform-family data are retained, not only model and role.

ClausHolbechArista: fixed

gmuloc (approved):

carl-baillargeon (approved):

PR #1753, opened 2026-09-15
fix: Handle absent gNSI service fields on older EOS

### LINKED ISSUE #1744: gNSI service facts treat unsupported fields as missing on older EOS
### Description

The gNSI service facts treat service fields omitted by older EOS response schemas as missing data. On releases where the service does not exist, the command itself can still succeed and return otherwise valid JSON.

This affects at least `GnsiAcctzFact` and `GnsiPathzFact` in `anta/_advisory/facts/management.py`.

### Reproduction

On EOS 4.31.10M and 4.31.2F:

```eos
show management api gnsi | json
```

returns:

```json
{
  "authzEnabled": false,
  "authzPolicy": false,
  "certzEnabled": false,
  "credentialzEnabled": false,
  "transports": {}
}
```

On EOS 4.32.0.3F and 4.32.1F the same command returns:

```json
{
  "authzEnabled": false,
  "authzPolicy": true,
  "certzEnabled": false,
  "credentialzEnabled": false,
  "transports": {}
}
```

In both response variants, `acctzEnabled` and `pathzEnabled` are absent. Attempting to enable the newer service syntax on these older releases also demonstrates the version boundary. For example:

```eos
management api gnsi
   service acctz
```

is rejected with an invalid-input error on these releases.

Newer EOS schemas include both fields, for example:

```json
{
  "acctzEnabled": false,
  "authzEnabled": false,
  "certzEnabled": false,
  "credentialzEnabled": false,
  "pathzEnabled": false,
  "transports": {}
}
```

### Current behavior

`GnsiAcctzFact` and `GnsiPathzFact` call `.get()` for their field and return `FactProblemKind.MISSING` when it is absent. This makes an unsupported subservice look like incomplete or malformed collection.

### Expected behavior

Recognize the valid older schema and normalize a version-absent gNSI subservice as `FeatureState.UNSUPPORTED`, or introduce another explicit version-aware representation. Missing fields in a schema that should contain them should remain distinguishable from an unsupported service.

### Impact

This is primarily an **indeterminate/error result**, not a false positive or false negative.

- For `GnsiAcctzFact`, SA174 can report Error instead of Not Affected on an affected older EOS release when P4Runtime is enabled with mTLS, P4Runtime request accounting is disabled, and Acctz is unavailable because EOS does not implement it. Treating the absent Acctz service as unsupported/disabled would allow the accounting path to be evaluated definitively.
- The observed Pathz omissions are on EOS 4.31 and 4.32, which are outside SA164's current affected-version matrix. They therefore do not currently change SA164's final result, although the standalone fact is still incorrect and unsafe to reuse without version context.

Suggested priority: low to moderate. It creates unnecessary errors and weakens fact reuse, but the observed cases do not directly report a vulnerable system as safe.

carl-baillargeon:
## Description

Make the gNSI Acctz and Pathz facts interpret absent service fields according to the EOS response-schema boundary:

- before EOS 4.33.2, normalize absent fields as unsupported services
- on EOS 4.33.2 and later, preserve absent fields as missing/incomplete schema data
- preserve malformed handling for fields present with invalid values and missing handling when EOS version metadata is unavailable

Fixes #1744

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing gNSI Acctz and Pathz fields based on EOS version.
  * Older EOS versions now report unsupported fields appropriately.
  * Newer EOS versions preserve missing-field results instead of incorrectly reporting malformed data.
  * Explicitly null Acctz and Pathz values are handled as missing while retaining validation for invalid boolean values.


anta/_advisory/facts/management.py
```diff
@@ -143,6 +150,32 @@ def _feature_source(command: AntaCommand) -> FactSource:
     return FactSource(command.command, FactSourceKind.COMMAND)
 
 
+def _version_aware_gnsi_service_absence(
```
gmuloc, line 153: for this PR I think it is ok but I would rather want a mechanism where we can say for a fact? a command? a list of train in which they have been introduced (not sure how exactly) so that we can make this automatically. Maybe let's wait for more and track this?

carl-baillargeon: Yes agreed: https://github.com/aristanetworks/anta/issues/1766.

gmuloc (approved):

PR #1758, opened 2026-09-16
fix: Treat non-running gNPSI transports as ineffective

### LINKED ISSUE #1745: gNPSI facts report non-running transports as effective
### Description

The gNPSI facts use configured enablement without considering whether the transport is running or has configuration errors. EOS can report both the service and transport as enabled while the effective transport is unusable.

This affects `GnpsiTransportFact`, `GnpsiAuthenticationExposureFact`, and `GnpsiMutualTlsSpiffeMitigationFact` in `anta/_advisory/facts/management.py`.

### Reproduction

On EOS 4.34.2.2F, 4.34.8M, and 4.35.0.2F, apply:

```eos
sflow run
management api gnpsi
   transport grpc default
      source sflow
      port 9339
      ssl profile ARISTA_DEFAULT_SELF_SIGNED_PROFILE
      authentication username priority metadata
      no disabled
```

`show management api gnpsi | json` returns:

```json
{
  "enabled": true,
  "transports": {
    "default": {
      "enabled": true,
      "running": false,
      "port": 0,
      "securityType": "unknown",
      "sslProfile": "",
      "authnUsernamePriority": [],
      "configErrors": {
        "noSslProfileFound": true
      }
    }
  }
}
```

A custom mTLS profile that is not usable by gNPSI produces the same effective state with:

```json
"configErrors": {"invalidSslProfile": true}
```

### Current behavior

- `GnpsiTransportFact` reports enabled from only the top-level `enabled` flag.
- The authentication and SPIFFE mitigation facts inspect the configured nested transport, encounter `securityType: "unknown"`, and return unavailable/malformed.

### Expected behavior

Effective gNPSI state should account for at least the transport `running` flag and `configErrors`. A configured but non-running transport with configuration errors should not be reported as an enabled exposure. Authentication and mitigation parsing should ignore or short-circuit ineffective transports instead of treating their intentionally unavailable runtime authentication fields as malformed JSON.

### Impact

At the fact level this is a **false positive**: `GnpsiTransportFact` says the transport is enabled/effective even though EOS says it is not running and has configuration errors.

For the observed output, SA158 does not currently produce a false Affected result because the authentication and mitigation facts become unavailable when `securityType` is `unknown`. Instead, an affected EOS release can produce Error where the non-running transport should make the result Not Affected. This is less severe than a false negative, but it creates an unnecessary operational alert and prevents a definitive assessment.

There is also a latent false-positive risk if EOS retains recognizable authentication fields while a transport is non-running: the current facts do not use `running` or `configErrors` as a gate and could then report an exposure that cannot accept requests.

Suggested priority: moderate. The demonstrated outcome is a false error, with a plausible false-Affected path for other valid EOS response combinations.

carl-baillargeon:
## Description

gNPSI facts previously treated configured enablement as effective runtime state. EOS can report the service and transport as enabled while `running` is false and `configErrors` is set (`noSslProfileFound` or `invalidSslProfile`), including `securityType: "unknown"`.

That made `GnpsiTransportFact` look enabled and the authentication/SPIFFE facts treat the unused transport as malformed JSON. SA158 then errored on an affected EOS release instead of returning Not Affected.

Effective transports now require configured enablement without `running: false` or a truthy configuration error. Authentication and mitigation facts inspect only that set. A missing `running` field stays a candidate to avoid false negatives; a candidate without a valid `configErrors` mapping is malformed.

Fixes #1745

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Security advisory evaluations now consider whether gNPSI transports are actively running and free of configuration errors, rather than relying only on enabled status.
  * Configured but non-running or errored transports are correctly treated as ineffective, preventing inaccurate exposure or mitigation findings.
  * Active, valid transports remain decisive even when another configured transport is unavailable.
  * Invalid runtime and configuration data is now rejected as malformed.


anta/_advisory/facts/management.py
```diff
@@ -852,13 +896,13 @@ def parse(cls, commands: tuple[AntaCommand, ...]) -> Fact[FeatureValue]:  # noqa
             return cls.unavailable(FactProblemKind.MALFORMED, source)
         if not config.enabled:
             return cls.available(FeatureValue(feature, FeatureState.DISABLED), source)
-        if any(not isinstance(transport.enabled, bool) for transport in config.transports):
-            return cls.unavailable(FactProblemKind.MALFORMED, source)
-        enabled_transports = tuple(transport for transport in config.transports if transport.enabled is True)
-        if not enabled_transports:
+        effective_transports = _effective_gnpsi_transports(config.transports)
```
ClausHolbechArista, line 899: clearly we are misusing the fact here. The fact should be a fact and the assessment decide if we are affected.

carl-baillargeon: I don't follow. Can you elaborate?

If there is no transport running the fact should be disabled so not affected right? Right now a non-running transport can be malformed.

PR #1765, opened 2026-09-17
fix: Assess SA158 credential logging from metadata authentication

### LINKED ISSUE #1746: GnpsiEosRpcAuthTraceFact cannot observe enabled trace state
### Description

`GnpsiEosRpcAuthTraceFact` cannot observe an explicitly enabled EosRpcAuth trace using its current command, `show trace Gnpsi | grep Auth`.

### Reproduction

On EOS 4.34.2.2F through 4.36.2F, configure gNPSI and enable the trace facility:

```eos
sflow run
management api gnpsi
   transport grpc default
      source sflow
      port 9339
      ssl profile ARISTA_DEFAULT_SELF_SIGNED_PROFILE
      no disabled
exit
exit
trace Gnpsi setting EosRpcAuth/4
trace Gnpsi enable EosRpcAuth all
```

The configuration commands succeed. However:

```eos
show trace Gnpsi | grep Auth
```

returns an empty successful response:

```json
{"output": ""}
```

This was consistent on 4.34.2.2F, 4.34.8M, 4.35.0.2F, 4.35.1F, 4.35.2F, 4.35.3F, 4.35.6M, 4.36.0.1F, 4.36.1F, and 4.36.2F.

### Current behavior

The fact interprets empty output as `FeatureState.DISABLED`, producing a false negative for explicitly enabled tracing.

### Expected behavior

Use an EOS command or output pattern that reliably exposes the configured EosRpcAuth trace state. If EOS does not provide a reliable query, the fact should return unavailable rather than claiming the trace is disabled.

### Impact

This is a **false negative** and is the highest-impact issue in this group.

SA158 checks the trace fact before the transport and EOS-version facts. Because empty output is normalized as disabled, an affected device with explicitly enabled EosRpcAuth tracing is reported Not Affected for CVE-2026-73457. The user receives no error or indication that the trace state was unobservable.

Until a reliable command is identified, returning unavailable/error for empty output is safer than returning disabled because it prevents vulnerable trace configuration from being silently classified as safe.

Suggested priority: high because this can hide an affected configuration.

carl-baillargeon:
## Description

SA158 treated empty `show trace Gnpsi | grep Auth` as EosRpcAuth disabled. That command can succeed with no output until the Gnpsi agent has handled gNPSI work, even when tracing is enabled. A subscription that later drops can still leave the facility visible, so empty output is not a reliable disabled signal and a currently active subscription is not required for output to appear. Timed exec can also enable EosRpcAuth later without a durable, reliably queryable trace state.

CVE-2026-73457 now uses metadata authentication as the logging prerequisite. On an affected release, enabled gNPSI with metadata authentication is Affected. Exclusive mTLS with only x509-spiffe is Not Affected, not Mitigated: the password-disclosure path is absent. Per the advisory assessment decision, absence of a necessary vulnerable feature is Not Affected even when the advisory text calls that configuration a mitigation. EosRpcAuth trace state is no longer collected.

CVE-2026-73456 is unchanged: TLS with metadata or mTLS with x509-common-name.

Fixes #1746

## Checklist


gmuloc:
discussion tracking:

`show trace Gnpsi | grep Auth `

empty output => cannot conclude
output present => potentially can conclude affected

anta/tests/advisories/sa_158.py
```diff
@@ -152,13 +138,16 @@ class SA158(OptionalCommandsMixin, _AntaAdvisoryTest):
...
+    * Success: EOS is outside scope, gNPSI is disabled, or the issue-specific authentication prerequisite is absent.
+    * Failure: An affected EOS release has an enabled gNPSI transport and the issue-specific authentication prerequisite.
+    * Error: Required EOS, gNPSI transport, or authentication state cannot be determined.
 
-    CVE-2026-73456 evaluates TLS or mTLS authentication combinations. CVE-2026-73457 evaluates explicit EosRpcAuth tracing and the
-    source-defined mutual-TLS/x509-spiffe mitigation across every enabled transport.
+    CVE-2026-73456 evaluates TLS metadata and mTLS common-name authentication. CVE-2026-73457 evaluates metadata
+    authentication on an affected release. Exclusive mTLS with only x509-spiffe is not affected: the password-disclosure
+    path is absent, which the assessment decision treats as not affected rather than mitigated. EosRpcAuth trace state
+    is not used: ``show trace Gnpsi`` can be empty until the agent has handled gNPSI work, and it can still report the
```
gmuloc: presence or absence of trace is not a good enough indicator of vulnerability as it is possible to activate the trace using timed exec ... at any time

carl-baillargeon: Updated. Please check.

